PR #18, opened 2019-04-05
Re-order blog posts

cbrevik:
This will make the newest blog post render on top when window width is larger than 1100px


mikaelbr:
Herlig! Kan merges for min del! 🆗

PR #47, opened 2019-12-16
Keypad instead of keyboard on iOS

jberglund:
iOS will now display numpad instead of keyboard. I haven't been able to test on Android due to lack of said phone.


adriansberg:
Works on Android with my third-party keyboard at least! ✅

mikaelbr:
Also verified #45

PR #61, opened 2020-08-11
Variant.no v2

mikaelbr:
All work on v2 of the new variant.no with updated profile.


src/app.css
```diff
@@ -0,0 +1,22 @@
...
+
+@import "~@variant/profile/colors/index.module.css";
+@import "~@variant/profile/typography/index.module.css";
+
+#root {
+  position: relative;
+}
+
+html {
+  background: var(--color-secondary4__tint2);
```
adriansberg, line 21: Any reason to separate this rule from the other rule at the top of the file?

mikaelbr: Not really, but just separating the "global general configuration" from the project specific. And in order for some specificity from the profile/color included above.

adriansberg: I see 👌

mikaelbr:
@MathiasFl I think actually `AZURE_PROXY_URL` and `AZURE_PROXY_KEY` should be public now that we limit the data. It's essentially just a JSON representation of what we already have. And it makes it sooo much easier to test it out locally.

mikaelbr:
This is merged now! Good job everyone.

PR #70, opened 2020-08-13
Adds varianttur subpage

mikaelbr:
Direct copy/paste with minor adjustments to have Varianttur-subpage with the new style.


src/varianttur/index.tsx
```diff
@@ -0,0 +1,257 @@
+import React, { useState } from "react";
+import style from "./varianttur.module.css";
+import Head from "next/head";
+
+const and = (...classes: string[]) => classes.join(" ");
```
adriansberg, line 5: I can see this useful as a utils function. Where should utility functions live, `src/utils`?

mikaelbr: Sure. Could move it as an util already. Usually, I move them in when they are needed (or even sometimes just duplicate, as it allows for easier deletion/cleanup and at low cost for such small helper functions). I think if we want them as utils we can do `src/utils/somenamehere.ts`. Don't know the name, though. But `react` maybe? At some point we could even collect `@variant/helpers`.

src/varianttur/varianttur.module.css
```diff
@@ -0,0 +1,212 @@
+:root {
+  --font-size-min: 12;
+  --font-size-max: 22;
+  --viewport-min: 400;
+  --viewport-max: 1500;
+}
+
+html {
```
adriansberg, line 8: I didn't think Next.js allowed global classes in module files, do we have config that allows it or is it always allowed?

mikaelbr: It does when you override the css loader 😬 

From the next config:
```
module.exports = withPlugins([withTM, withImages], {
  target: "serverless",
  webpack: (config) => {
    const oneOf = config.module.rules.find(
      (rule) => typeof rule.oneOf === "object"
    );
    if (oneOf) {
      const moduleCssRule = oneOf.oneOf.find(
        (rule) => regexEqual(rule.test, /\.module\.css$/)
        // regexEqual(rule.test, /\.module\.(scss|sass)$/)
      );
      if (moduleCssRule) {
        const cssLoader = moduleCssRule.use.find(({ loader }) =>
          loader.includes("css-loader")
        );
        if (cssLoader) {
          cssLoader.options.modules.mode = "local";
        }
      }
    }
    return config;
  },
});
```

☝️ Finds files that are CSS modules and set module mode to local for the CSS loader.

adriansberg: Just found that code 😁 Does that mean dl/dd/dt for instance only affects the `varianttur` page?

mikaelbr: Yes, it should only be included for that specific page.

PR #72, opened 2020-08-13
Adds footer to layout

mikaelbr:
Adds a footer to the layout:

image

Might be we need to restructure/make a separate "clean" layout for pages that don't need a footer.


adriansberg:
The spacing between `footer__item`s (decided by `min-width: 15rem;`) feels a bit off balance to me. 

Something like this looks better IMO (just a grid with gap).
image

MathiasFl:
The desktop version looks good:+1:. On mobile view the footer svg is to steep, and creates a large void before next valuable content. The pairing wave is also missing but that may be intentional. The alignment of the text blocks is also a little wonky vertically, when they become fully stacked. But i like the layout on wider mobiles where we can fit 2 horizontal (See 2nd image).

Screenshot 2020-08-14 at 08 06 29
Screenshot 2020-08-14 at 08 13 58

mikaelbr:
True! I was thinking about this before sleep yesterday, I have a solution in mind that does not require break points

PR #74, opened 2020-08-14
Adds silly animation on mouse movement for footer

mikaelbr:
It's very subtle and not very exciting at all, but it is some movement. Obviously (now) it's not very exciting with parallax on such a simple and two-dimensional illustration. Hard to fake 3d using parallax when they are positioned like this.

But it is some sort of movement which makes it a bit more intriguing. Albeit not much.


src/layout/layout.module.css
```diff
@@ -63,7 +63,16 @@
 }
 
 .footer {
-  background: url("./footer.svg") no-repeat center top;
+  position: relative;
+  overflow: hidden;
+  /* background: url("./footer.svg") no-repeat center top; */
```
adriansberg, line 68: Unused now?

mikaelbr: Good catch! 🚀

src/layout/footer-image/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        height="2392"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+        className={style.svg}
+      >
+        <path
+          fillRule="evenodd"
+          clipRule="evenodd"
+          d="M1682 183.658l60-24.038c60-24.038 180-72.114 300-66.105 120 6.01 240 66.105 360 78.124 120 12.019 240-24.038 360-18.029 120 6.01 240 54.086 300 78.124l60 24.038V370H1682V183.658z"
+          fill="#E61A6B"
```
adriansberg, line 37: These colors can be substitued with colors from @variant/profile?

mikaelbr: You are right, my friend! Fixed

PR #78, opened 2020-08-15
ci: Adds eslint for hooks only and github actions for validating

mikaelbr:


.github/workflows/eslint.yml
```diff
@@ -0,0 +1,19 @@
+name: Lint
+
+on:
```
cbrevik, line 3: File/glob filter here?

.github/workflows/tsc.yml
```diff
@@ -0,0 +1,23 @@
+name: TypeScript
+
+on:
+  push:
+    branches:
```
cbrevik, line 5: Same?

.github/workflows/eslint.yml
mikaelbr: Good point. Also, don't really need the pull_request trigger when having push. Can just have push on all branches with a path and it will work out. hmm. Unless external pull requests? Hm? How would that work out, do you know?

mikaelbr: Fixing it if it's an issue. Just merging. YOLO

PR #79, opened 2020-08-16
feat: Add employees view 

### LINKED ISSUE #60: Ansattvisning
Lag en visning som viser alle ansatte

adriansberg:
Initial implementation of the employee view. Have to look into why all the blobs end up looking the same regardless of different seeds.

Todo:
- [ ] Fix blobs so all aren't the same blob
- [ ] Use a blob for the links to jobs.variant.no on the employees page
- [ ] Do scaling of images properly
- [ ] Should there be a link in the nav to the employees page? 

Closes #60


mikaelbr:
Fantastic! Some quick notes and thoughts while multitasking:

- Yes, should be link in the header. 😃
- i dont think the header title should be graphik but rather recoletta.
- i think there is a too big size difference between name and phone?
- i think it's too much space between name and image.
- random order would be nice 😀
- some render issue on home page while on mobile
- would be cool having a filter/search option. Also "im feeling lucky" to pick a random one

adriansberg:
> Fantastic! Some quick notes and thoughts while multitasking:
> 
>     * Yes, should be link in the header. 😃
> 
>     * i dont think the header title should be graphik but rather recoletta.
> 
>     * i think there is a too big size difference between name and phone?
> 
>     * i think it's too much space between name and image.
> 
>     * random order would be nice 😀
> 
>     * some render issue on home page while on mobile
> 
>     * would be cool having a filter/search option. Also "im feeling lucky" to pick a random one

Sizes and spaces are taken from Figma. I very much agree that the phone number should be bigger. I can tweak the sizes and come up with a proposal. On that note, is there an overview of the sizes used? Do you have any experience with any filter library as [fuzzy](https://www.npmjs.com/package/fuzzy) or the like?

adriansberg:
> Regarding filtrering. Wouldn't it be cool if one could filter based on some extra key words : react Java recruiting design etc? Odd Morten Sveås
> […](#)
> 17. aug. 2020 kl. 16:27 skrev Mikael Brevik ***@***.***>: ﻿ Fantastic! Some quick notes and thoughts while multitasking: Yes, should be link in the header. 😃 i dont think the header title should be graphik but rather recoletta. i think there is a too big size difference between name and phone? i think it's too much space between name and image. random order would be nice 😀 some render issue on home page while on mobile would be cool having a filter/search option. Also "im feeling lucky" to pick a random one — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Should these keywords be saved directly in the json file that contains the rest of the information?

adriansberg:
Should there be a link to the employees view in the navbar?

mikaelbr:
@adriansberg See my first comment :D Yes, add a link.

Also, regarding @oddsve's comment. I think we should use CV Partner as source. I can create an open API in the styleguide repo for the blob implementation which we can use. From that we can get keyword searches, filtering on names, and even images. This means we don't have to manually maintain an employee list.

adriansberg:
> @adriansberg See my first comment :D Yes, add a link.
> 
> Also, regarding @oddsve's comment. I think we should use CV Partner as source. I can create an open API in the styleguide repo for the blob implementation which we can use. From that we can get keyword searches, filtering on names, and even images. This means we don't have to manually maintain an employee list.

That would be _really_ good!

mikaelbr:
But I don't think the API should be a blocker for this, though. It might take some time to do the OAuth flow, so I think a hard coded JSON will do just as fine for the release of this.

adriansberg:
We can do basic filtering now and expand it with CV Partner later on.

mikaelbr:
Didn't answer your questions from before. Sorry! Yeah, I've used fuzzy, and other similar implementations. It works as expected. For our use case, we have such a small data set that it is almost no performance reduction at all to do fuzzy matching (both for user patterns and CPU cycles), so I think using fuzzy is fine.

AFAIK, the spaces/sizes in Figma aren't too polished/perfectly planned (please correct me @jberglund, @hildeor or @pettersuul ). So I think we can adjust it some. At least propose changes. The sizes I've used is a multiple of 20. So for the site it self it is usually 2.4 and 4.8. And for the typography, it's based on the style guide typo which is well defined.

oddsve:
Also, there could be good to have control of what key words that are to be exposed. I think we could use only approved competencies, but I recon there are other key words as well. I hard coded JSON sounds good as a first attempt.

adriansberg:
I get a mismatch between the server and the client for the path of blobs. Any idea how to resolve?

`Warning: Prop `d` did not match.`

mikaelbr:
If setting the same seed doesn't do it (haven't checked the code), you have the option of suppressing the warning (https://reactjs.org/docs/dom-elements.html#suppresshydrationwarning). As far as i can tell it wouldn't cause any big side effects just ignoring it

PR #90, opened 2020-08-23
V2 cv partner proxy

MathiasFl:


functions/.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
mikaelbr, line 1: Should we have .vscode in both root and this folder?

functions/getEmployees/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      // Make images
+      const employeeList = employeesJSON.filter(
+        (employee) =>
+          employee.deactivated ||
+          // All employees that have started should be set
+          // as countrymanager in cv-parter, when they start.
+          employee.roles.some((role) => role === "countrymanager")
+      );
+      context.res = {
+        body: employeeList,
```
mikaelbr, line 36: Should we limit the amount of data we pass on to just what we need? In this case just the images. Both to limit data so we don't expose things we don't intend to. But also to reduce payloads.

functions/getEmployees/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      };
+    } else {
+      context.res = {
+        status: 404,
+        body: await request.clone(),
```
mikaelbr, line 41: Wow. Did not know about `.clone` 🤯

functions/getEmployees/sample.dat
```diff
@@ -0,0 +1,3 @@
+{
```
mikaelbr, line 1: Is this needed?

functions/.vscode/settings.json
MathiasFl: Good catch, was auto generated i have ```.vscode``` hidden in my explorer.

mikaelbr (approved):

functions/getEmployees/sample.dat
MathiasFl: Probably not, i could make proper dummy data. But i have never experienced someone using it 🤔

mikaelbr:
Should we ignore `functions/` from vercel? https://vercel.com/guides/prevent-uploading-sourcepaths-with-vercelignore

functions/getEmployees/index.ts
MathiasFl: It's nice, but actually not needed here 🤦. Was a previous problem that needed `.clone` this to work.

functions/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "functions",
```
mikaelbr, line 2: Could name this something else and mark it as `"private": true` just for good measure.

functions/getEmployees/index.ts
MathiasFl: Had the same thought, but came to the conclusion that it would be easier to just add an extra layer of data massage. Excess data are handled during the `.getStaticProps` so user should never be able to access that data we don't want to exopse (unless we start to use the api client-side). Are we paying for bandwith on Vecrel?, on Azure it's runtime so its cheaper to not do extra operations on the data.

mikaelbr: Not paying for bandwidth. I think the minor transformation cost on CPU is neglectable for functions also, tbh. This is fairly theoretical and somewhat loaded of paranoia, but given that you expose all data, you could technically push a PR passing on the entire data set on a branch and log that. I don't think it is likely though, but yeah.

MathiasFl:
Added a `.vecrelignore` for `functions` so we save some time and space when deploying.

mikaelbr:
Paranoia

functions/getEmployees/index.ts
MathiasFl: I understand what you mean, and maybe was part of the reason for moving it behind a proxy in the first place.

mikaelbr:
Coolbeans! Is this deployed on push through Azure now?

MathiasFl:
Yes, and .env variables has been added to preview in Vecrel.

mikaelbr:
Fantastic. Merge at will for what I'm conserned 👍

PR #92, opened 2020-08-26
Navbar mag

Dahly96:
Made a more responsive header where it creates a burger menu and adds 'Si Hallo' to the menu instead. No styling added to the menu other the whats used on the text, hover and positioning.


src/layout/index.tsx
```diff
@@ -20,6 +20,23 @@ const Layout: React.FC<LayoutProps> = ({
 }) => {
   const footerContainer = useRef<HTMLElement>(null);
 
+  const [clickActive, setClickActive] = useState(true);
```
mikaelbr, line 23: I wonder if you can do the same in pure CSS and avoid using JavaScript implemented break points as a whole.

Haven't tested out, but something similar to:
1. Have menu as position absolute with proper styling on break-point in CSS. Normal style if not.
1. Button visible on some breakpoint
1. "Si hallo!" not visible at the opposite breakpoint.

having both "hamburger button and "Si hallo" position absolute makes them not take up any spacing in the layout.

mikaelbr: And also have a click-toggle state for visible/not visible for the menu. When breakpoint max-width something the visible (display: block or whatever) trumphs the toggle state.

mikaelbr:
I've added a comment on how I think it can be simplified without using JavaScript for listening on resize. Other than that it seems to work great 👌  Some styling on the menu itself and we're golden

mikaelbr:
Have you gotten any design feedback on this, yet?

src/layout/index.tsx
```diff
@@ -51,8 +68,52 @@ const Layout: React.FC<LayoutProps> = ({
...
+              onClick={() =>
+                clickActive ? setClickActive(false) : setClickActive(true)
+              }
+            >
+              <div
+                className={and(
+                  style.bar1,
+                  clickActive ? style.bar1_change : '',
+                )}
+              ></div>
```
adriansberg, line 96: These tags can be self-closed.

src/layout/layout.module.css
```diff
@@ -19,22 +21,37 @@
...
 }
+
+.header__nav__ul li:last-child {
+  display: none;
+}
+
+.header__nav__ul a {
+  font-family: Recoleta;
+  color: #333333;
+  font-size: 14px;
```
adriansberg, line 52: Should use `rem`s. Not for the benefit of relative sizing, but more for consistency throughout the code.

src/layout/layout.module.css
```diff
@@ -19,22 +21,37 @@
...
+  transition: 0.4s;
 }
+
+.header__nav__ul li:last-child {
+  display: none;
+}
+
+.header__nav__ul a {
+  font-family: Recoleta;
+  color: #333333;
```
adriansberg, line 51: Should use color from @variant/profile

src/layout/layout.module.css
```diff
@@ -57,6 +74,96 @@
...
+
+.bar1_change,
+.bar2_change,
+.bar3_change {
+  position: relative;
+  width: 25px;
+  height: 3px;
+  background-color: var(--color-secondary1);
+  margin: 4px 0;
+  transition: 0.4s;
```
adriansberg, line 113: Can use CSS-modules' `composes` to prevent duplicate code. 

`composes: bar1`

adriansberg (changes requested):
Some quick remarks:

- The button for closing the menu disappears when scrolling down the page
- Maybe there could be some slide in animation or something when opening the menu
- The links could be larger
- Just an idea: Instead of the menu being a square container, maybe it could have a more wavy shape like the footer (or blobby if you like)?

Dahly96:
> Have you gotten any design feedback on this, yet?

Yes, have implemented the design inputs!

src/layout/layout.module.css
Dahly96: done^^

src/layout/index.tsx
Dahly96: fixed :D

src/layout/layout.module.css
Dahly96: fixed :D

src/layout/layout.module.css
Dahly96: fixed :D

src/layout/index.tsx
Dahly96: fixed :D

src/layout/burger-menu/burger-blob.svg
```diff
@@ -0,0 +1,10 @@
+
```
mikaelbr, line 1: This looks to be unused now?

src/layout/index.tsx
```diff
@@ -20,8 +20,28 @@ const Layout: React.FC<LayoutProps> = ({
 }) => {
   const footerContainer = useRef<HTMLElement>(null);
 
+  const [clickActive, setClickActive] = useState(false);
+  const burgerRef = useRef<HTMLDivElement>(null);
+
+  const handleClick = (e: Event) => {
```
mikaelbr, line 26: This could have a more specific name. Such as `onClickOutside` or something.

src/layout/index.tsx
```diff
@@ -20,8 +20,28 @@ const Layout: React.FC<LayoutProps> = ({
...
+    if (burgerRef.current && !burgerRef.current.contains(e.target as Node)) {
+      setClickActive(false);
+    }
+  };
+
+  useEffect(() => {
```
mikaelbr, line 32: Thinking more about this, I think actually all this `handleClick` now with `nav` as 100% width and height all you need is an `onClick` listener on the nav element through React. That would render the useEffect obsolete.

mikaelbr: Also `burgerRef`

src/layout/index.tsx
```diff
@@ -51,30 +71,95 @@ const Layout: React.FC<LayoutProps> = ({
...
+          </span>
+
+          <div ref={burgerRef}>
+            <button
+              className={style.container}
+              id="hamburger"
+              aria-labelledby="menu-label"
+              aria-expanded={clickActive ? true : false}
+              onClick={() =>
+                clickActive ? setClickActive(false) : setClickActive(true)
```
mikaelbr, line 86: This can simpler be expressed as `onClick={() => setClickActive(!clickActive)}` as you want to toggle clickActive between true/false (which is the same as negating the current value).

src/layout/index.tsx
```diff
@@ -51,30 +71,95 @@ const Layout: React.FC<LayoutProps> = ({
...
+          <span hidden id="menu-label">
+            Hovedmeny
+          </span>
+
+          <div ref={burgerRef}>
+            <button
+              className={style.container}
+              id="hamburger"
+              aria-labelledby="menu-label"
+              aria-expanded={clickActive ? true : false}
```
mikaelbr, line 84: This can be simpler as `aria-expanded={clickActive}`

src/layout/index.tsx
```diff
@@ -51,30 +71,95 @@ const Layout: React.FC<LayoutProps> = ({
...
+              />
+            </button>
+
+            <nav
+              className={and(
+                style.header__nav,
+                clickActive ? '' : style.header__nav__hidden,
+              )}
+              aria-labelledby="menu-label"
+              aria-hidden={clickActive ? true : false}
```
mikaelbr, line 115:
```suggestion
              aria-hidden={clickActive}
```

src/layout/index.tsx
mikaelbr:
```suggestion
              aria-expanded={clickActive}
```

src/layout/index.tsx
mikaelbr:
```suggestion
                setClickActive(!clickActive)
```

src/layout/index.tsx
```diff
@@ -51,30 +71,95 @@ const Layout: React.FC<LayoutProps> = ({
...
+              id="menu"
+            >
+              <ul className={style.header__nav__ul} id="nav-ul">
+                <li>
+                  <a
+                    href="https://jobs.variant.no"
+                    rel="noopener"
+                    tabIndex={clickActive ? 0 : -1}
```
mikaelbr, line 123: Hmm. What is this for as it is now? It causes the links to not be keyboard accessible on desktop.

src/layout/layout.module.css
```diff
@@ -57,6 +68,113 @@
 .header__nav a:hover {
   color: rgba(0, 0, 0, 0);
 }
+
+.container {
```
mikaelbr, line 72: Think this can be named something more specific. Like `buttonContainer`

mikaelbr (approved):
🚀 Some suggestions of minor things and fixes that could be done. I think removing buttonRef and useEffect is a quick win to reduce unnecessary complexity.

src/layout/burger-menu/burger-blob.svg
Dahly96: yeah, its removed now. Going to go over and clean up a bit of code!

src/layout/index.tsx
Dahly96: Yes, thats right! thanks for the input :D

src/layout/index.tsx
Dahly96: Yes, thats changed ^^

src/layout/index.tsx
Dahly96: When i think about it, if clickActive is true then hidden should be false right?

src/layout/index.tsx
Dahly96: its there so that you cant tab to it when its not shown. Before i put that inn you could tab to the sidebar even if it was hidden to the side, it was accessible either way. When its -1 it disables tabbing to the links in the <ul.

src/layout/layout.module.css
Dahly96: agreed!

src/layout/index.tsx
Dahly96: Based on the comment bellow i fixed it with just an onClick and removed burgerRef and useEffect.

Dahly96: also changed the name to onClickOutsideUl

src/layout/index.tsx
mikaelbr: There are cases now on desktop where you cannot tab the visible menu in the header.

Dahly96: I will look into that!

src/layout/index.tsx
Dahly96: Yes! made it into a onclick function instead!

src/layout/index.tsx
Dahly96: Should be okey now, going to push soon. made a few changes

src/layout/index.tsx
Dahly96: changed it to aria-hidden={!clickActive} ^^

src/layout/layout.module.css
adriansberg: Can be even more specific IMO.

src/index/index.tsx
```diff
@@ -28,6 +28,7 @@ const Home = ({
           content="En ny variant av et konsulentselskap som er raust, åpent og modig. Vi bistår med utvikling, design, salg, kursing og rådgiving."
         />
       </Head>
+
```
mikaelbr, line 31:
```suggestion
```

mikaelbr:
This works 👍  I think we are good to merge for now

mikaelbr:
@Dahly96 Feel free to merge

PR #97, opened 2020-09-02
Projectref view

Dahly96:
Added two new references and made it so 3 refs is random generated and displayed on the page. Only need pictures for the two new ones now.


pages/index.tsx
```diff
@@ -2,11 +2,18 @@ import { GetStaticProps } from 'next';
 import { handleImages } from 'src/utils/imageHandler';
 import { Employee, massageEmployee } from 'src/employees';
 import { getEmployeesUrl } from 'src/utils/api/getEmployees';
+import { CaseJSON } from '../src/case/Case';
```
mikaelbr, line 5: In `tsconfig.json` I've added path aliasing as an experiment to seeing if it is better to use absolute pathing instead of relative pathing with `../../` etc. Not as problematic here, but it's nice to be consistent.

pages/index.tsx
```diff
@@ -2,11 +2,18 @@ import { GetStaticProps } from 'next';
...
+import { CaseList } from '../src/case/cases';
 
 export { default } from 'src/index';
 
+function shuffle(array: Array<CaseJSON>) {
```
mikaelbr, line 10: Here you only use it for one thing here, so not necessary, but I think this is a perfect use case for [generics](https://www.typescriptlang.org/docs/handbook/generics.html). Example:

```
function shuffle<T>(array: Array<T>) { }
```

pages/index.tsx
```diff
@@ -2,11 +2,18 @@ import { GetStaticProps } from 'next';
...
+  array.sort(() => Math.random() - 0.5);
```
mikaelbr, line 11: Sort mutates, but it also returns. To ensure the original won't be "ruined" you can copy it before. It's not uncommon to see something like 

```
  [...array].sort(() => Math.random() - 0.5);
```

pages/index.tsx
```diff
@@ -16,7 +23,12 @@ export const getStaticProps: GetStaticProps<{
       employeesJSON[Math.floor(Math.random() * employeesJSON.length)];
     const imageSlug = await handleImages(employee);
     const randomEmployee = { ...massageEmployee(employee), imageSlug };
-    return { props: { randomEmployee }, revalidate: 24 * 60 * 60 };
+
+    const getCases: CaseJSON[] = CaseList;
+    shuffle(getCases);
+    const randomCases = getCases.slice(0, 3);
```
mikaelbr, line 29: Given the `[...array]` from the previous comment this can be written as

```
const randomCases = suffle(CaseList).slice(0, 3);
```

mikaelbr (approved):

pages/index.tsx
Dahly96: okey nice :D changing now^^

pages/index.tsx
Dahly96: Yes, going over it now and chaging :D

pages/index.tsx
Dahly96: good call, going over it now :D

src/utils/api/getEmployees.ts
```diff
@@ -1,5 +1,5 @@
 const externalGetEmployeesURL =
-  'https://variant-homepage-proxy.azurewebsites.net/api/getEmployees?code=rM%2FOUPE7QtnGa6el%2FAro%2FT2u7Zc029LU2CaOtlbn6lz4l3UoceauVw%3D%3D';
+  'https://variant-homepage-proxy.azurewebsites.net/api/getEmployees';
```
mikaelbr, line 2: Is this a regression error now? cc @MathiasFl

Dahly96: Ahh yeah there was some issue while testing. But it might need to be put back before merge

MathiasFl: Apparently tries Azure to authenticate you with passed key even if auth is set to "anonymous", so when the missing app keys are happening it's not possible to test. This was temp hotfix, should be removed from commit or get updated by rebasing onto master before merge.

mikaelbr (approved):
LGTM

PR #98, opened 2020-09-03
Removes some unnecessary/outdated code

mikaelbr:
- Blobs are no longer animated in any way, so removing lottie/bodymovin is a quick win.
- Also removes animation from footer as it is unnecessary on not really pretty.
- Updates dependencies to styleguide (getting new font styles as provided by @jberglund )


jberglund (commented):
👍

PR #99, opened 2020-09-10
add custom wrapper to prevent rendering on server.

MathiasFl:
Issue with react-slider and serverside rendering. 
image

Add a component wrapper to avoid rendering on server.


mikaelbr:
Could we also just do `dynamic` here?

mikaelbr (approved):
Looks good! Think this is a better solution

mikaelbr:
Denne fungerer vel ikke uten JavaScript uansett, så det med årstallet er kanskje ikke så krise?

PR #100, opened 2020-09-12
fix(ansatte): Fix slicing of phone numbers

adriansberg:


src/employees/index.tsx
```diff
@@ -202,7 +202,7 @@ export const massageEmployee = (
     fullName: employee.name,
     name: employee.name.split(' ')[0],
     phone: (employee.telephone.startsWith('+47')
-      ? employee.telephone.slice(2)
+      ? employee.telephone.slice(3)
```
mikaelbr, line 205: Might at some point be more sophisticated by using regex as `00NN`, `+NN` is valid also. For instance something like `.replace(/^(00|\+)\d\d/, '')`

mikaelbr (approved):

src/employees/index.tsx
adriansberg: Definitely!

PR #109, opened 2020-12-10
fix: Adds bug fixes for menu and font size.

mikaelbr:
- Should fix issue with menu being off when navigating back in Safari.
- Should also add better keyboard navigation when navigation is visible and scrollable area
- And lastly, reduces max size of font by upgrading @varianer/profile dependency where it has been fixed permanently.

### DIFF
--- a/src/layout/index.tsx
+++ b/src/layout/index.tsx
@@ -19,12 +19,10 @@ const Layout: React.FC<LayoutProps> = ({
 }) => {
   const modalRef = React.createRef<HTMLDivElement>();
   const navRef = React.createRef<HTMLUListElement>();
-  const {
-    isMenuVisible,
-    setMenuVisible,
-    tabIndex,
-    // isNotHamburgerMode,
-  } = useTogglableBurgerMenu(modalRef, navRef);
+  const { isMenuVisible, setMenuVisible, tabIndex } = useTogglableBurgerMenu(
+    modalRef,
+    navRef,
+  );
 
   return (
     <div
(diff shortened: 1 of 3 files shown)


src/layout/index.tsx
```diff
@@ -17,43 +17,19 @@ const Layout: React.FC<LayoutProps> = ({
...
-      return -1;
-    }
-  }
+  const modalRef = React.createRef<HTMLDivElement>();
+  const navRef = React.createRef<HTMLUListElement>();
+  const {
+    isMenuVisible,
+    setMenuVisible,
+    tabIndex,
+    // isNotHamburgerMode,
```
adriansberg, line 26: Remove?

src/layout/index.tsx
```diff
@@ -269,3 +248,92 @@ const Layout: React.FC<LayoutProps> = ({
...
+        return e.preventDefault();
+      }
+
+      // On "reversed" tabbing. If previous element is outside modal, jump to last element
+      if (e.shiftKey && !modalRef.current?.contains(previous)) {
+        last?.focus();
+        return e.preventDefault();
+      }
+
+      // Normal tab flow.
```
adriansberg, line 314: Misplaced comment or just me not understanding?

mikaelbr: Can specify better! Point is, no case hits, so just follow normal browser behaviour.

adriansberg: Right, gotcha.

adriansberg:
I can't close the menu with the cross 😳

mikaelbr:
Good catch! Fixed now

adriansberg (approved):

PR #111, opened 2020-12-16
Changes to new CV images with optimized images

mikaelbr:
- Adds lazy loading webp support
- Removes saving 150/300 as we handle that through next/image
- No longer doing blobification.


mikaelbr:
When lazy loading there is no placeholder which can make it seem odd. But I don't think it is the biggest issue.

PR #112, opened 2020-12-17
Accessibility improvements

jberglund:
Added linting for as a step towards better accessibility. A tiny step, but a step nonetheless. 

If you've got ESlint running in VSCode, you'll hopefully get errors when you're about to do something that ain't proper for devices such as screen readers.

In this PR:
- The GitHub action for linting was outdated. I changed it accordingly.
- Linting will now occur by the grace of Githubus Actionatius XIV on push to this repo instead of pre-commit.
- I took the liberty of removing eslint from staged-lint. It wasn't running anyway because it was set to only lint `*.js`. Of which there are none!
- Added `*.tsx, *.ts` to the prettier pre-commit hook. This is the way.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,7 @@ module.exports = {
     es2020: true,
     node: true,
   },
-  extends: ['plugin:react-hooks/recommended'],
+  extends: ['plugin:react-hooks/recommended', 'plugin:jsx-a11y/recommended'],
   parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaFeatures: {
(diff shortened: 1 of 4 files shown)


mikaelbr:
I totally forgot when chatting with you previously btw. But there is accessibility checking before, although apparently not as strict as eslint jsx-a11y. But we have a bot running: AccessLint

jberglund:
> I totally forgot when chatting with you previously btw. But there is accessibility checking before, although apparently not as strict as eslint jsx-a11y. But we have a bot running: AccessLint

Have you ever seen it report anything? Asking because curious. 🤔 
Lets disable the Github Action, but keep the a11y plugins. Agreed?

PR #114, opened 2021-02-05
fix: issue with tab navigation when menu is not in burger mode

mikaelbr:


src/layout/index.tsx
```diff
@@ -346,11 +349,55 @@ function useTogglableBurgerMenu<
...
 }
+
+function hasWindow() {
+  return typeof window !== 'undefined';
+}
+
+const useMediaQuery = (mediaQuery: string) => {
+  const [isMatched, setMatched] = useState(() => {
+    if (!hasWindow()) return false;
+    return Boolean(window?.matchMedia(mediaQuery).matches);
```
adriansberg, line 368: `hasWindow()` doesn't remove the need of using a questionmark here?

adriansberg (approved):
Looks good to me 🚀

PR #162, opened 2021-09-09
Sommersplash 2021

toretefre:
Work in progress, not to be reviewed or merged until all mentioned issues are resolved.

This pull requests brings a splash that should be visible on variant.no/sommerjobb and variant.no/nyutdannet beginning before ITdagene starts the 19th of September, until the application deadline for graduates in Variant the 3rd of October. 
The commits should probably be squashed before merge.

Good stuff:
- The splash is (mostly) implemented as intended

Bad stuff: 
- Snap scroll is not implemented, due to issues believed to be related to the fixed layer with changing background colors. Without snap scroll it does not make sense to have page / slide counter, so this was removed.

Issues that needs to be resolved before production:
- [x] Files that are not needed in /splashlayout and /summersplash should be removed before merge
- [x] Z-index issue making active tab in modeTabs section being visible above fullWidth menu

Screenshots:
image

image


toretefre:
Sett kort gjennom sammen med Marius

PR #166, opened 2021-09-17
Improve responsiveness on summersplash

toretefre:
Tested with devtools for iPhone X, iPad, 1440x900 and on MBP 16

Also mobile layout for graduate-calculator:

image


mikaelbr (approved):
Ser bra ut på mobil i alle fall! Kjør på

PR #176, opened 2021-11-29
Change employee filter to use active harvest users

mathiasflaatt:
Change filter logic on the function that fetches all employees to use active users in respective Harvest, instead of random role in CV partner.

Due to missing access to other offices in harvest, I can only verify it works for Trondheim.

The function is deployed on its own deployment slot "harvesttest" so the valid URL for this PR is https://variant-homepage-proxy-harvesttest.azurewebsites.net/api/getemployees.  (Url is updated for this Preview in Vercel)


mikaelbr:
@nikolaia du har adgang til å generere keys for Harvest Oslo?

mathiasflaatt:
Ser ut som det kan brukes en API token for alle account-ids, hvis det er noen som har den tilgangen. Slik som det er gjort på [create-variant-user repoet](https://github.com/varianter/create-variant-user/blob/a08b01c8877af0ecaaf84604ca293b99c80f9eea/src/App.js#L63)

PR #233, opened 2022-08-02
fix: #229 download of podcast episode now starts when you press play

mortennordseth:
fix: #229 by setting the preload attribute of the audio element to none. The podcast duration that was previously fetched using `preload="metadata"` is added in the mapFeedToPodcast function.


cbrevik (approved):
LGTM! Bra jobba 👨🏻‍🎤

PR #259, opened 2022-08-25
fix: listing of multiple departments

mikaelbr:
Adds support for having multiple departments.

Still not the best handling here as it compares unstructered data. Somewhat more laxing with case insensitive regex, but still 😅


trulshj (approved):

trulshj:
Nice one!

PR #311, opened 2022-10-21
feat(sec): add security headers

42tte:
About Nextjs security headers https://nextjs.org/docs/advanced-features/security-headers
About CSP https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -14,6 +14,58 @@ const regexEqual = (x, y) => {
   );
 };
 
+const ContentSecurityPolicy = `
+  default-src 'self';
+  connect-src 'self' https://variant.innocraft.cloud/ https://g.nav.no/api/v1/;
+  script-src 'self' 'sha256-j6xN8x073Dhm+Ee4HKwIIRXsHIqI5aIRHC0pgnhVcJY=' https://variant.innocraft.cloud/ ${
+    process.env.NODE_ENV !== 'production' ? "'unsafe-eval'" : ''
+  };
+  style-src 'self' 'unsafe-inline' http://hello.myfonts.net/;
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -14,6 +14,57 @@ const regexEqual = (x, y) => {
   );
 };
 
+const ContentSecurityPolicy = `
+  default-src 'self';
+  connect-src 'self' https://variant.innocraft.cloud/;
+  script-src 'self' 'sha256-j6xN8x073Dhm+Ee4HKwIIRXsHIqI5aIRHC0pgnhVcJY=' https://variant.innocraft.cloud/ ${
+    process.env.NODE_ENV !== 'production' ? "'unsafe-eval'" : ''
```
42tte, line 21: Needed for hot reload in dev build

next.config.js
```diff
@@ -14,6 +14,57 @@ const regexEqual = (x, y) => {
   );
 };
 
+const ContentSecurityPolicy = `
+  default-src 'self';
+  connect-src 'self' https://variant.innocraft.cloud/;
+  script-src 'self' 'sha256-j6xN8x073Dhm+Ee4HKwIIRXsHIqI5aIRHC0pgnhVcJY=' https://variant.innocraft.cloud/ ${
```
42tte, line 20: Should have used 'strict-dynamic' but there is no good way of using generated nonces in Nextjs https://github.com/vercel/next.js/discussions/21587

mikaelbr:
Haven't gone through all potential external references here, but at a glance it looks to cover the things I can think of at least. Now there will be a new implementation of variant webpage, so this would have to be ported over. In that port I also think we could transition to using middleware which I belive now can support nonce.

mikaelbr (approved):

PR #312, opened 2022-10-29
Calculate experience

oddsve:
I think comparing to a typical grad date makes more sense. It´s not always correct, I know, but most young people graduated in before summer.


oddsve:
selvsagt basert på feil branch

PR #315, opened 2022-11-04
42tte/fix a11y

42tte:


src/rss/feed/Blog.tsx
```diff
@@ -6,7 +6,7 @@ import { truncateOnSpace } from './utils';
 
 export default function Blog({ item }: { item: BlogItem }) {
   return (
-    <div className={and(style.card, style['card--blog'])}>
+    <a className={and(style.card, style['card--blog'])}  href={item.url} target="_blank" rel="noopener noreferrer">
```
mikaelbr, line 9: Not harming anyone to include it, but I actually think noopener is default on _blank links in most/all browsers as per spec now (https://html.spec.whatwg.org/#following-hyperlinks)

src/utils/use-media-query.ts
```diff
@@ -1,45 +1,27 @@
...
+  }, []);
 
   useEffect(() => {
-    if (!hasWindow()) return;
-    const mediaQueryList = window.matchMedia(mediaQuery);
-    const documentChangeHandler = () =>
-      setMatched(Boolean(mediaQueryList.matches));
-    listenTo(mediaQueryList, documentChangeHandler);
+    const media = window.matchMedia(query);
+    media.addEventListener('change', updateTarget);
```
mikaelbr, line 16: Safari did/didn't support addEventListener on MediaQueryList due to an overly strict following of [the spec](https://drafts.csswg.org/cssom-view/#the-mediaquerylist-interface). Have you verified that this is no longer the case in most used Safari versions?

mikaelbr: Yes, seems to be updated in Safari 14
Screenshot 2022-11-06 at 20 01 58
Screenshot 2022-11-06 at 20 00 33

src/advanced-calculator/calculator.module.css
```diff
@@ -97,12 +102,16 @@ input[type='number'] {
...
   width: 100%;
-  position: sticky;
   top: 0;
   grid-column: 2 / 3;
   grid-row: span 1;
   padding: 1.5rem;
 }
+@media (min-width: 900px) {
+  .calculator-controls {
+    position: sticky;
```
mikaelbr, line 112: Strictly speaking not an a11y change IMO. Can do this change but then I'd argue that we can remove the show/hide toggle also as it does nothing.

src/advanced-calculator/index.tsx
```diff
@@ -435,7 +435,7 @@ export default function Calculator() {
             </ul>
           </InView>
           <InView>
-            <h5>Familie</h5>
+            <h4 className="var-heading var-heading--5">Familie</h4>
```
mikaelbr, line 438: Feel like this at some point could be it's own title + level + override element component that can be reused on different systems. Now we've been mixing style and semantics

src/components/button/index.tsx
```diff
@@ -50,10 +50,10 @@ export function ButtonNextLink({
   className = '',
   children,
   ...props
-}: ButtonProps & LinkType) {
+}: ButtonProps & LinkType & Pick<AType, 'aria-label'>) {
   return (
     <Link {...props}>
-      <a className={and(style.buttonLink, className)}>{children}</a>
+      <a className={and(style.buttonLink, className)} aria-label={props["aria-label"]}>{children}</a>
```
mikaelbr, line 56: This this might rather be rewritten to use `<Link>` without `<a>` child. Current version of `next/link` doesn't require `<a>`-child as the previous versions did. Would simplify types given that `Link` supports aria attrs

mikaelbr (approved):
Good job and nice improvements. 🚀  

I still think if not controls on the calculator are sticky it shouldn't be toggleable. Also if it isn't sticky we should add some numbers where comparisons are necessary. For cases on pension, they should see what we offer and when it is not sticky they would have to scroll to the top to be reminded. I'd argue that this is a far more important feature than having more screen space in landscape mode. Might be a design flaw but I think not being ideal in landscape mode is the lesser of two evils.

src/rss/feed/Blog.tsx
42tte: Resolved

src/advanced-calculator/index.tsx
42tte: Created a Heading component for reuse

src/components/button/index.tsx
42tte: That would have been nice, [next/link](https://nextjs.org/docs/api-reference/next/link) don't have className so ended up keeping this as is

src/advanced-calculator/calculator.module.css
42tte: [This is an a11y issue](https://www.uutilsynet.no/wcag-standarden/134-visningsretning-niva-aa/141). Not current law, and won't apply to us as a private company (yet). But still have effect on real people.
Can agree the solution could be better but that would have to be figured out.

mikaelbr: To be clearer, as mentioned in a previous comment. A lot of what we do here is trying to clearly show potential salery and in some cases comparing numbers. And on phones this is if I recall correctly the only place it is shown. So with this change we scroll down and the essential information is no longer easily accessible. My opinion is that that has more negative impact in the short run than taking too much space in landscape. As it affects more users. Non is ideal ofc. Fixing the underlying issue with the design (or extending to be more sophisticated with in/out by scroll etc) is best but probably more time consuming. My opinion is that until that is prioritized the current non-ideal affects less people than this proposed non-ideal.

42tte: Dropped for now. Should I create an Issue or how should we address this?

mikaelbr: Might also have @vegardfeste for instance look at it to see if there's a quick win here, but at least solve it in a separate issue so we can merge this.

42tte: Everything should be good, will you resolve comments

PR #418, opened 2023-10-20
feat: adds link to variant.dev in footer

mikaelbr:
Adds link to variant.dev in footer, feed and læreglede section


mikaelbr:
YOLO

PR #439, opened 2024-03-20
fix: return 404 for unpublished jobs and adds preview url

mikaelbr:
Hides old job listings, but make them available under `jobs/preview/[slug]`


andreashartveit:
Knall i padden 🐸 Dette er nice!

PR #461, opened 2024-08-13
Added PR-template

anemne:


mathiazom (approved):
😎

christinaroise:
I wonder if we should add a note to specify which "environment" we've added changes to.. Because we have studio, web and storybook.. 🤔 Might just end up adding noise though.

anemne:
> I wonder if we should add a note to specify which "environment" we've added changes to.. Because we have studio, web and storybook.. 🤔 Might just end up adding noise though.

What about adding "breaking changes"? I suggest that we keep it as simple as possible.

PR #466, opened 2024-08-14
Initial calculator page + ErrorMessage

mathiazom:
(resurrected #462)

Added dummy `Calculator` component in a new page at `/kalkulator`.

image

Also replace `ErrorNews` with a more univeral `ErrorMessage` component

image

(original `ErrorNews` as a special case of `ErrorMessage`)

image

Finally fixed some minor issues encountered when using a fresh dataset.


src/app/(main)/page.tsx
```diff
@@ -34,7 +35,7 @@ const Home = async () => {
   );
 
   if (!landingId) {
-    return <NewTemplatePlaceholder />;
+    return <MissingData description={"Landing page id"} />
```
christinaroise, line 38: we can do this in a separate PR and not something pressing!

we know that if landingId doesn't exist it's most likely because sanity studio is empty. In terms of UX we should add a nicer Welcome screen with some description on how to get started. I can create something in figma for us to use as a template and create an issue based off of that.

src/app/(main)/page.tsx
```diff
@@ -44,7 +45,11 @@ const Home = async () => {
   );
 
   if (!initialLandingPage) {
-    throw new Error("Page not found");
+    return <MissingData description={`Page for id '${landingId}'`} />
+  }
+
+  if (!initialLandingPage.data.sections) {
+    return <MissingData description={`Sections for landing page`} />
```
christinaroise, line 52: for later: I think this should have the same welcome screen later so that the if would be ` if(!initialLandingPage || !initialLandingPage.data.sections)`

src/calculator/calculator.module.css
```diff
@@ -0,0 +1,9 @@
+.content {
+    margin-top: 10rem;
```
christinaroise, line 2: why do you still need the margin-top?

src/styles/global.css
```diff
@@ -12,6 +12,7 @@ html {
   --primary-green-dark: #206a6f;
   --primary-green-light: #a9db91;
   --primary-red-error: #b30b0b;
+  --primary-red-error-light: #ffebee;
```
christinaroise, line 15: let's not add new colors. 

I think you can use the same color schema as here for the missingData.module. That way it's easier to refactor once we get the proper design
Screenshot 2024-08-14 at 14 06 11

christinaroise (approved):
Left a few comments, some relevant some others. Adjust accordingly 🎉

src/calculator/calculator.module.css
mathiazom: The calculator will be pushed all the way to the top otherwise. 

image

I think this is the case for alle the pages, but they have enough padding/margin to not be visible.

image

src/components/errorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,67 @@
...
+  _type: 'link',
+  linkTitle: 'Go to Studio',
+  linkType: LinkType.Internal,
+  internalLink: {
+    _ref: 'studio',
+  },
+};
+
+
+export const MissingContentErrorMessage = ({ description }: { description: string }) => {
```
christinaroise, line 32: Hm, this is a good setup but it's also easy to get confused as the current structure groups components into unique folders:  errorMessage/ErrorMessage. 

So when seeing MissingContentErrorMessage I'd expect it to be in: missingContentErrorMessage/MissingContentErrorMessage

What we can do is set ErrorMessage with a type og better yet make it more universal like:

```
const ErrorMessage = ({ onClick }: { onClick?: () => void }) => {
  return (
    <section className={styles.wrapper}>
      <div className={styles.error}>
        <Text type="h3">{title}</Text>
        <Text>{description}</Text>
        <div className={styles.buttonWrapper}>
          {onClick && (
            <Button type="secondary" onClick={onClick}>
              Try again
            </Button>
          )}
          <LinkButton type="secondary" link={link} />
        </div>
      </div>
    </section>
  );
};
```

that way we make sure that error messages are universal in it's content and only have button as optional

src/calculator/calculator.module.css
christinaroise: you can add padding then. I know we'll omit this component later, but let's go with the structure that has been set previously to not add confusion as to how the code is structured.

what I typically do is look at section elements and try to adapt the wrapper styling — that's typically what design does as well.

src/components/errorMessage/ErrorMessage.tsx
mathiazom: So the "Return to home" link should be present in all cases?

mathiazom: But we can definitely move them into separate folder if that's cleaner

image

mathiazom: Also, having a `description` only as a string (instead of allowing a children node) will limit what you can show in the error message. That migh be fine depending on the use cases.

christinaroise: you can set link as optional also if that's better; but for cases with !landingID I'd set the link as Go to Studio and then path /studio

mathiazom: Updated now to include optional link and tryAgain callback in generic component

src/components/errorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,28 @@
+import Text from 'src/components/text/Text';
```
christinaroise, line 1: nitpicky about this, sorry.

omit children, we don't want that because that omits the purpose of universal components and instead creates wrappers for static components. 

what I want is ErrorMessage as the only error component:

```
import Button from "src/components/buttons/Button";
import LinkButton from "src/components/linkButton/LinkButton";
import Text from "src/components/text/Text";
import { ILink } from "studio/lib/payloads/navigation";
import styles from "./errorNews.module.css";

// custom because ErrorMessage exists natively
interface CustomErrorMessageProps {
  title: string;
  description: string;
  onClick?: () => void;
  link?: ILink;
}

const CustomErrorMessage = ({
  title,
  description,
  onClick,
  link,
}: CustomErrorMessageProps) => {
  return (
    <section className={styles.wrapper}>
      <div className={styles.error}>
        <Text type="h3">{title}</Text>
        <Text>{description}</Text>
        <div className={styles.buttonWrapper}>
          {onClick && (
            <Button type="secondary" onClick={onClick}>
              Try again
            </Button>
          )}
          {link && <LinkButton type="secondary" link={link} />}
        </div>
      </div>
    </section>
  );
};

export default CustomErrorMessage;
```

 
 that way we can replace NewsError like this in News.tsx:
```
   if (!initialPosts || posts.error) {
    const link = {
      _key: "return-home",
      _type: "link",
      linkTitle: "Return to home",
      linkType: LinkType.Internal,
      internalLink: {
        _ref: "/",
      },
    };

    return (
      <CustomErrorMessage
        title={`Error - Can’t fetch news`}
        description="Struggling to fetch stories, and we dont know why. Please be patient
          and try to refresh the page."
        onClick={fetchPosts}
        link={link}
      />
    );
  }
```


and MissingContentErrorMessage in (main)/page.tsx would be:

```
const studioLink = {
  _key: "go-to-studio",
  _type: "link",
  linkTitle: "Go to studio",
  linkType: LinkType.Internal,
  internalLink: {
    _ref: "/studio",
  },
};

const Home = async () => {
  const { perspective, isDraftMode } = getDraftModeInfo();

  const { data: landingId } = await loadQuery<string>(
    LANDING_QUERY,
    {},
    { perspective }
  );

  if (!landingId) {
    return (
      <CustomErrorMessage
        title="Landing Page Not Set"
        description="It looks like no page has been designated as your landing page in the Studio. Head over to the Studio to set a landing page and guide visitors to the right place!"
        link={studioLink}
      />
    );
  }

  const initialLandingPage = await loadQuery<PageBuilder>(
    PAGE_QUERY,
    { id: landingId },
    { perspective }
  );

  if (!initialLandingPage) {
    return (
      <CustomErrorMessage
        title="Landing Page is Missing Content"
        description={`Your landing page is set, but it looks like there’s no content yet. Visit the Studio to start adding content and make your landing page come to life!`}
        link={studioLink}
      />
    );
  }
///



```

mathiazom: We can omit children. That would only limit what we can put into the error message, which is probably fine for this use case.

However, having only one component would require duplication of common props within groups of errors. For example the studio link and text strings that should be common for all errors relating to missing content. That might not be a big problem for this specific component, so I can change it to one component for now, but I think it would be limiting if applied as a general rule of the codebase.

christinaroise: I disagree. If you need to duplicate link or string you can keep these as export constants in CustomErrorMessage.tsx

What you can also do is create a type for CustomErrorMessage: `type ErrorType = "missing content" | "default"` to be used in place of link and onClick in cases such as missing content. 

⚠️🚨  Ops, I've also forgot to add: `role="alert" aria-live="assertive"` to the error message for a11y ⚠️🚨

mathiazom:
@christinaroise, should I change this PR to only include the error fixes and dummy calculator? Error message is now covered by #474

christinaroise:
Hm, yes, let's make this a bit cleaner and only include what's related to calc. (Error cases are handled in separate PRs)

mathiazom:
Moved calc to #479

PR #470, opened 2024-08-16
feature(radiobuttongroup): added radiobutton and radiobuttongroup

anemne:
This is just an example of how the radioButtons will look like now: 
Screenshot 2024-08-19 at 15 57 47

The design will be updated once the sketches are completed. I've removed all content from the salary and benefits form, so this is just a placeholder for visualization purposes.


src/app/calculator/page.tsx
```diff
@@ -0,0 +1,30 @@
+"use client";
```
christinaroise, line 1: make sure this isn't merged to v3 as it's just for testing that the radio group works. 💡

src/components/buttons/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+  name,
+  value,
+  label,
+  checked,
+  disabled,
+  defaultChecked,
+  onChange,
+}) => {
+  return (
+    <div aria-label={label}
```
christinaroise, line 28: why are you setting the aria-label here?

src/components/buttons/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+  checked,
+  disabled,
+  defaultChecked,
+  onChange,
+}) => {
+  return (
+    <div
+      className={`${styles.main}`}
+    >   
+    <input
```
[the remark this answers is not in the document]
christinaroise: hm I don't think the lint-test is correct as it doesn't recognises htmlFor

src/components/buttons/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+        className={`${styles.input}`}
+        type="radio"
+        name={name}
+        id={id}
+        value={value}
+        checked={checked}
+        disabled={disabled}
```
christinaroise, line 38: let's look at disabled and aria-disabled, we might actually want to use aria disabled and then add extra js logic to make sure it's function is also disabled. 

with disabled the element is hidden from a11y tech, so a screenreader wouldn't read the element/option. but with aria-diabled it'd read it, but list it as disabled, meaning it's not hidden.

src/components/buttons/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,47 @@
+"use client";
```
christinaroise, line 1: I don't think you need use client for this component?

src/components/buttons/RadioButton/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+    label, 
+    options,
+    onChange
+}) => {
+    return (
+        <fieldset className={styles.fieldset} id={id}>
+            <legend className={styles.legend}>{label}</legend>
+            <div className={styles.wrapper}> 
+            <RenderOptions options={options} onChange={onChange} />
+                {/* <RadioButton key="radio-one" label="Bachelor" id="radio-one" name="radio" disabled={false}/>
```
christinaroise, line 30: omit the comment?

src/components/buttons/RadioButton/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,35 @@
+"use client";
```
christinaroise, line 1: omit use client as there's no useState e.i

src/components/buttons/RadioButton/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+    id: string;
+    label: string;
+    disabled: boolean;
+}
+
+interface RadioButtonGroupProps {
+    id: string;
+    label: string;
+    options: IOption[];
+    onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
```
christinaroise, line 17: I'm missing selected or a default selected, ref: https://medium.com/@christinaroise/how-to-create-a-reusable-custom-radiobutton-in-react-with-typescript-3ae7fc704e09

With radio groups there has to be one option pre-selected. 

We can set this in two ways:
- add an option prop: selectedOption?: IOption;
- set selected option to selectedOption || options[0];

src/components/buttons/RadioButton/RenderOptions.tsx
```diff
@@ -0,0 +1,35 @@
+"use client";
```
christinaroise, line 1: I think you can add this inside RadioButtonGroup.tsx but outside RadioButtonGroup, so:

```
const Options = () => {};
export const RadioButtonGroup = () => {};
```

src/components/buttons/RadioButton/radioButton.module.css
```diff
@@ -0,0 +1,68 @@
+.main {
```
christinaroise, line 1: I wonder if we should structure our module.css for typography differently. 

Perhaps we should create label classes in text.module.css and then use these here. We can add a note about this in readme so that we make sure that text.module.css is the only global module.css file.

christinaroise:
Nice! I'm starting to think that the [accesslint[bot]](https://github.com/apps/accesslint) doesn't take htmlFor etc into account, we'll have to take a look at how to update the bot or omit that rule set for inputs

src/components/buttons/RadioButton/RadioButton.tsx
anemne: I'll look into it!

src/components/forms/radioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,96 @@
...
+ *   label="Choose an option"
+ *   options={options}
+ *   onValueChange={(name, value) => console.log(`Selected ${name}: ${value}`)}
+ * />
+ * ```
+ * - In this example, the `RadioButtonGroup` component renders multiple `RadioButton` components 
+ *   as part of a cohesive group, enabling proper radio button functionality.
+ */
+
+export const RadioButtonGroup: React.FC<RadioButtonGroupProps> = ({
```
christinaroise, line 58:
```suggestion
export const RadioButtonGroup = ({
    id,
    label, 
    options,
    onValueChange}: 
    RadioButtonGroupProps)
```

src/components/forms/radioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <fieldset className={styles.fieldset} id={id}>
+            <legend className={textStyles.h3}>{label}</legend>
+            <div className={styles.wrapper}> 
+                <RenderOptions options={options} onChange={handleChange} />
+            </div> 
+        </fieldset>  
+    );
+};
+
+const RenderOptions: React.FC<RenderOptionsProps> = ({ options, onChange }) => {
```
christinaroise, line 79:
```suggestion
const RenderOptions  = ({ options, onChange }: RenderOptionsProps) => {
```

src/components/forms/radioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <>
+            {options.map(({ id, label, disabled }, index) => (
+                <RadioButton
+                    key={id}
+                    id={id}
+                    label={label}
+                    name="radio"
+                    disabled={disabled}
+                    value={label}
+                    defaultChecked={index === 0}
```
christinaroise, line 90:
```suggestion
                    defaultChecked={currentValue}
```

src/components/forms/radioButtonGroup/components/RadioButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+  name: string;
+  label: string;
+  value?: string;
+  checked?: boolean;
+  disabled?: boolean;
+  defaultChecked?: boolean;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+export const RadioButton: React.FC<RadioButtonProps> = ({
```
christinaroise, line 16:
```suggestion
export const RadioButton = ({
  id,
  name,
  value,
  label,
  checked,
  disabled,
  defaultChecked,
  onChange,
 }: RadioButtonProps
```

src/components/forms/radioButtonGroup/RadioButtonGroup.tsx
christinaroise: or currentChecked 🤷‍♀️  We want the selectedOption to come from parent, say the form or page

src/components/forms/radioButtonGroup/radioButtonGroup.module.css
```diff
@@ -0,0 +1,62 @@
+.container {
```
christinaroise, line 1: split what's related to radiobuttongroup and radiobutton in separate css as we now have RadioButton under components 💡

src/components/forms/radioButtonGroup/components/RadioButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+  name,
+  value,
+  label,
+  checked,
+  disabled,
+  defaultChecked,
+  onChange,
+}) => {
+  return (
+    <>
```
christinaroise, line 27: we can remove fragments as they're now redundant
```suggestion

```

src/components/forms/radioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,96 @@
...
+interface RenderOptionsProps {
+    options: IOption[];
+    onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+interface RadioButtonGroupProps {
+    id: string;
+    label: string;
+    options: IOption[];
+    onValueChange: (name: string, value: string) => void;
```
christinaroise, line 21: Normally, we'd set:
```suggestion
    onValueChange: (option: IOption) => void;
```
given that options are IOption[]

src/components/forms/radioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        onValueChange(e.target.name, e.target.value);
+    };
+
+    return (
+        <fieldset className={styles.fieldset} id={id}>
+            <legend className={textStyles.h3}>{label}</legend>
+            <div className={styles.wrapper}> 
+                <RenderOptions options={options} onChange={handleChange} />
```
christinaroise, line 73:
```suggestion
                <RenderOptions options={options} onChange={() => onValueChange(option)} />
```


not entirely sure.. 

Also, maybe we should test and see how RenderOptions render differently from {options.map()}?

christinaroise (approved):

PR #471, opened 2024-08-16
v3 - Fix/hero styling for firefox

christinaroise:
This fixes render of styling on firefox of CTAs in Hero section. 

Firefox:
Screenshot 2024-08-16 at 13 14 26

Safari: (buttons, from design, has a max-width. Safari doesnt scale smaller than this) 
Screenshot 2024-08-16 at 13 14 45

Chrome (same case here as with Safari, buttons have a max-width prop)
Screenshot 2024-08-16 at 13 15 53

### DIFF
--- a/src/components/linkButton/linkButton.module.css
+++ b/src/components/linkButton/linkButton.module.css
@@ -1,9 +1,11 @@
 .button {
-  width: -webkit-fill-available;
+  width: 100%;
+  box-sizing: border-box;
   max-width: 360px;
   cursor: pointer;
   display: inline-flex;
   align-items: center;
+  align-self: stretch;
   justify-content: space-between;
   gap: 0.6rem;
   border-radius: 1.5625rem;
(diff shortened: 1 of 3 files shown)


src/components/linkButton/linkButton.module.css
```diff
@@ -1,9 +1,11 @@
 .button {
-  width: -webkit-fill-available;
+  width: 100%;
+  box-sizing: border-box;
```
mathiazom, line 3: Have you considered adding `border-box` as a global default?

```
html {
  box-sizing: border-box;
}
*, *:before, *:after {
  box-sizing: inherit;
}
```
(https://css-tricks.com/box-sizing/#aa-universal-box-sizing-with-inheritance)

christinaroise: Good idea! I'll give it a test run to make sure it doesn't break any other components.

christinaroise: I'm adding this as a separate task/PR.

christinaroise: ...because it's more a global fix and nor only related to hero

anemne (approved):

mathiazom (approved):

PR #473, opened 2024-08-16
v3 — fix styling of testimonial image and checkbox on smaller screens

christinaroise:
this adds design updates to testimonials on small screens with rectangular images and the custom checkbox.

Screenshot 2024-08-16 at 17 27 52
Screenshot 2024-08-16 at 17 28 00


src/components/forms/checkbox/checkbox.module.css
```diff
@@ -4,7 +4,13 @@
   align-items: center;
   gap: 0.5rem;
   align-self: stretch;
-  width: -webkit-fill-available;
+  width: 100%;
+  color: var(--primary-black);
+
+  /* to keep checkbox 20x20 we'll make sure portable text takes 2/3 width */
+  & > p {
+    flex: 2;
```
christinaroise, line 12: not the best approach maybe, but chose to do it this way to avoid filling DOM with bloat, and to keep PortableText from style props.

mathiazom:
Would be nice to have "Before" images in the PR (especially when proper Sanity is not set up locally)

mathiazom:
The circular versions seem to have regressed a little with this change

| Before | After |
:-------------------------:|:-------------------------:
| image | image |

christinaroise:
> The circular versions seem to have regressed a little with this change

Good catch, I'll have to have a talk with the designer with that one as the padding was updated, but wouldn't really apply to much to us though 🤔

christinaroise:
@mathiazom a little mishap on padding. it was never suppose to change!

mathiazom (approved):

PR #474, opened 2024-08-16
v3 — Fix/error messages to pages

christinaroise:
This is built on @mathiazom [work ](https://github.com/varianter/variant.no/pull/466)

Converts ErrorNews to CustomErrorMessage and adds the message in app. places such as (main)/page.tsx and and Blog.tsx. (main)/[slug]/page.tsx has a redirect and will get a snackbar to alert visitors. 

Screenshot 2024-08-16 at 21 03 48

Screenshot 2024-08-16 at 21 04 48

Screenshot 2024-08-16 at 21 05 56


src/blog/components/customErrorMessage/CustomErrorMessage.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const CustomErrorMessage = ({
+  title,
+  body,
+  button,
+  link,
+}: CustomErrorMessageProps) => {
+  return (
+    <section className={styles.wrapper}>
```
christinaroise, line 42: might want to give this a test to check if we want it or not. 🤔 
`role="alert" aria-live="assertive"`

studio/lib/queries/navigation.ts
```diff
@@ -35,9 +35,9 @@ export const NAV_QUERY = groq`
         }
       }
     }
-  }[1]
+  }
```
mathiazom, line 38: This will return a list of navigation managers, breaking code expecting it to be a single object.
```suggestion
  }[0]
```

studio/lib/queries/siteSettings.ts
```diff
@@ -4,4 +4,4 @@ export const SITESETTINGS_QUERY = groq`*[_type == "siteSettings"]{
     brandAssets,
     siteMetadata,
     legalPages,
-}[0]`;
+}`;
```
mathiazom, line 7: Same as above

src/blog/components/customErrorMessage/CustomErrorMessage.tsx
mathiazom: Not sure of it fits this description

> `aria-live="assertive"` should only be used for time-sensitive/critical notifications that absolutely require the user's immediate attention. Generally, a change to an assertive live region will interrupt any announcement a screen reader is currently making. As such, it can be extremely annoying and disruptive and should only be used sparingly.

(https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#live_regions)

christinaroise: I think it does fit.

That way the focus is on the error message upon loading and not on "skip to main" and header. 💡

mathiazom: I guess it depends on how the component is used. If it's a full page error it probably fits, but if it's only concerning a small not-so-important component it might be annoying?

christinaroise: Regardless if it's a full page or a box, it's an error with a warning status, if it were an information box, it'd have aria-live="assertive".

In this case the ErrorMessage works as a full page.

studio/lib/queries/navigation.ts
christinaroise: I'm double checking.

Navigation manager along with a few other documents cannot be [] due to a custom structure (DeskStructure) so that might be why I've changed it. I'll have to look through and remember what I did on Friday 😅

src/blog/components/customErrorMessage/CustomErrorMessage.tsx
christinaroise: ops no, I meant to write polite 😂

studio/lib/queries/navigation.ts
christinaroise: Just me forgetting to push one commit. I've also added an extra check for proper render of different state changes in layout.

src/app/(main)/layout.tsx
mathiazom, line 1: Missing an import for `styles`

mathiazom (approved):

PR #475, opened 2024-08-16
v3 — Safely fetch navigationManager in Sanity

christinaroise:
With the newest configuration of environment labels, we have to adjust our api calls in custom sanity components. 

Sanity is run client-side, so to securely fetch data, we're introducing a utility function called `fetchWithToken`. This ensures that API tokens are not exposed client-side.

Key changes:
- Updated README
- Introduced a new API Route for Data Fetching (to be used with fetchWithToken)
- Implemented fetchWithToken Utility
- Updated CustomCallToActions Component to use fetchWithToken. Also added UI for loading and error states, as well as setting a "not available" message:
Screenshot 2024-08-16 at 20 25 36


src/app/api/navigationManager/route.ts
```diff
@@ -0,0 +1,32 @@
+// app/api/navigationManager/route.ts
+import { NextResponse } from "next/server";
+import { client } from "studio/lib/client";
+import { token } from "studio/lib/token";
+
+// Ensure the client is configured with the token
```
christinaroise, line 6: I think I want to expand on this comment, maybe add something to readme.

In order to fetch in custom sanity component we need the token, however the newest setup of the env ensures the token is only accessible server-side — but all sanity components run client side. This means that all calls inside sanity has to be created in the api-folder as their own routes.

studio/components/CustomCallToActions.tsx
```diff
@@ -3,45 +3,79 @@ import {
...
+
+    setIsLandingPage(isLanding);
+  }, [documentId, navigationManager]);
+
+  if (loading) {
+    return <LoadingBlock />;
+  }
+
+  if (error) {
+    return <div>Error: {error}</div>;
```
christinaroise, line 67: I wonder if we should do a card here like with !isLandingPage... 🤔

src/app/api/navigationManager/route.ts
christinaroise: We're going to have to do the same with most of the custom Sanity components. I'll give it some thought to see if there's a better approach to this.

studio/components/AnchorSelect.tsx
```diff
@@ -45,7 +45,9 @@ const AnchorSelect = ({ value, onChange, path }: AnchorSelectProps) => {
     }
 
     try {
-      const response = await client.fetch("*[_id == $id][0]{sections}", {
+      const response = await fetchWithToken<{
+        sections: AnchorItem[];
+      }>("*[_id == $id][0]{sections[]{_key, basicTitle}}", {
```
mathiazom, line 50: What is the difference between `sections[]` and `sections`?

studio/lib/fetchWithToken.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @throws Will throw an error if the fetch operation fails or if the response is not ok.
+ *
+ * @example
+ * const data = await fetchWithToken<MyDataType>("*[_type == 'myType']", { param1: 'value' });
+ */
+export const fetchWithToken = async <T>(
+  query: string,
+  params?: Record<string, any>
+): Promise<T> => {
+  const response = await fetch("/api/fetchData", {
```
mathiazom, line 25: Could we use [Server Actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) here?

christinaroise: hmm.. could be that I gave it a try only with client.fetch and not loadQuery! I'll give it a test run, cause it'd be nice to omit this as well! 🎉

christinaroise: ok so I looked into it, the reason why it's not working is because this is a custom React component within Sanity. and server actions, such as the "use server" directive, are specifically designed for use within Next.js server components, not client-side React components like the one we’re using in Sanity 💡

studio/components/AnchorSelect.tsx
mathiazom: Yes, that part was clear, but the extra `[]`  after `sections` seemed redundant. After double checking is seems to be required, so all good 🤗

studio/lib/fetchWithToken.ts
mathiazom: Makes sense if Sanity Studio is not using Next.js, so api route is the way to go here 👍 (but server actions can definitely be used in Next.js client components, so could be useful later)

mathiazom (approved):

PR #478, opened 2024-08-19
v3 — add margin-top to main global for offset sticky header

christinaroise:
So we don't have to add margin top to every case where header is not rendered, we're adding offsetForStickyHeader to main

### DIFF
--- a/src/app/(main)/layout.module.css
+++ b/src/app/(main)/layout.module.css
@@ -1,0 +1,3 @@
+.offsetForStickyHeader {
+  margin-top: 72px;
+}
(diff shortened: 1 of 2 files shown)


src/app/(main)/layout.module.css
```diff
@@ -0,0 +1,3 @@
+.offsetForStickyHeader {
```
mathiazom, line 1: Where is the class applied?

christinaroise: The changes was added to a diff PR so I have to adjust. give it two sec 👍

anemne (approved):

mathiazom (approved):

PR #484, opened 2024-08-20
v3 - benefit schema

### LINKED ISSUE #481: Add Benefit schema type

### LINKED ISSUE #483: Add Benefits document as Benefit array

mathiazom:
Closes #481

### DIFF
--- a/studio/schemas/documents/benefit.ts
+++ b/studio/schemas/documents/benefit.ts
@@ -1,0 +1,22 @@
+import { defineType } from 'sanity';
+import { richText, title } from '../fields/text';
+
+export const benefitId = "benefit";
+
+const benefit = defineType({
+  name: benefitId,
+  type: "document",
+  title: "Benefit",
+  fields: [
+    title,
+    richText
+  ],
+  preview: {
+    select: {
+      title: title.name,
+      subtitle: "Benefit"
+    }
+  }
+})
+
+export default benefit;


studio/schemas/documents/benefit.ts
```diff
@@ -0,0 +1,16 @@
+import { defineType } from 'sanity';
+import { richText, title } from '../fields/text';
+
+export const benefitId = "benefit";
+
+const benefit = defineType({
+  name: benefitId,
+  type: "document",
+  title: "Benefit",
```
christinaroise, line 9: How about adding a preview with title as title from fields and subtitle as Benefit? ⭐

studio/schemas/documents/benefit.ts
```diff
@@ -0,0 +1,16 @@
+import { defineType } from 'sanity';
```
christinaroise, line 1: Love it! But would only merge this file and omit the others

studio/schemas/deskStructure.ts
```diff
@@ -8,11 +8,13 @@ import {
   ControlsIcon,
```
christinaroise, line 8: revert as we don't want benefit to be a doc and this is only for preview.

studio/schema.ts
```diff
@@ -11,6 +11,7 @@ import blog from "./schemas/documents/blog";
 import posts from "./schemas/documents/post";
```
christinaroise, line 11: revert as we don't want benefit to be a doc and this is only for preview.

christinaroise (changes requested):

christinaroise (approved):

PR #492, opened 2024-08-20
v3 - salary and benefits page preview

mathiazom:
image

### DIFF
--- a/src/salaryAndBenefits/SalaryAndBenefitsPreview.tsx
+++ b/src/salaryAndBenefits/SalaryAndBenefitsPreview.tsx
@@ -1,28 +1,37 @@
-"use client"
+"use client";
 import { Suspense } from "react";
 import SalaryAndBenefits from "./SalaryAndBenefits";
-import { QueryResponseInitial, useQuery } from '@sanity/react-loader';
-import {SalaryAndBenefits as SalaryAndBenefitsPayload} from "studio/lib/payloads/salaryAndBenefits";
-import { SALARY_AND_BENEFITS_PAGE_QUERY } from 'studio/lib/queries/pages';
-
+import { QueryResponseInitial, useQuery } from "@sanity/react-loader";
+import { SalaryAndBenefits as SalaryAndBenefitsPayload } from "studio/lib/payloads/salaryAndBenefits";
+import { SALARY_AND_BENEFITS_PAGE_QUERY } from "studio/lib/queries/pages";
 
 interface SalaryAndBenefitsPreviewProps {
   initialSalaryAndBenefits: QueryResponseInitial<SalaryAndBenefitsPayload>;
 }
 
-const SalaryAndBenefitsPreview = ({initialSalaryAndBenefits}: SalaryAndBenefitsPreviewProps) => {
-
-  const {data: newData } = useQuery<SalaryAndBenefitsPayload>(
+const SalaryAndBenefitsPreview = ({
+  initialSalaryAndBenefits,
+}: SalaryAndBenefitsPreviewProps) => {
+  const { data: newData } = useQuery<SalaryAndBenefitsPayload>(
     SALARY_AND_BENEFITS_PAGE_QUERY,
     { slug: initialSalaryAndBenefits.data.slug.current },
-    { initial: initialSalaryAndBenefits }
+    { initial: initialSalaryAndBenefits },
   );
+
+  if (process.env.NODE_ENV !== "production") {
+    if (typeof newData !== "object" || newData === null) {
+      console.error(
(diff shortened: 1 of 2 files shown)


src/salaryAndBenefits/SalaryAndBenefitsPreview.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const SalaryAndBenefitsPreview = ({initialSalaryAndBenefits}: SalaryAndBenefitsPreviewProps) => {
+
+  const {data: newData } = useQuery<SalaryAndBenefitsPayload>(
+    SALARY_AND_BENEFITS_PAGE_QUERY,
+    { slug: initialSalaryAndBenefits.data.slug.current },
+    { initial: initialSalaryAndBenefits }
+  );
+
```
christinaroise, line 20: It doesn't affect performance necessarily, but in terms of code convention, maintainability and readability.. it's a safer option. we could go with `??` in stead of `||`, but since we know newData is a valid object or null/undefined it's sort of redundant.

```suggestion
const pageContent = newData || initialSalaryAndBenefits.data
```

src/salaryAndBenefits/SalaryAndBenefitsPreview.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Suspense>
+      <SalaryAndBenefits salaryAndBenefits={newData ?? initialSalaryAndBenefits.data} />
```
christinaroise, line 23:
```suggestion
      <SalaryAndBenefits salaryAndBenefits={pageData} />
```

christinaroise (approved):
approved 🔥  but add changes!

src/salaryAndBenefits/SalaryAndBenefitsPreview.tsx
mathiazom: Probably makes no practical difference here, but I usually prefer `??` over `||` when providing default values. `??` only returns the right-hand side if the left-hand side is `null` or `undefined`, while `||` returns the right-hand side if the left-hand side is any _falsy_ value. To me that is a much more vague operator, which could lead to unexpected results. You would for example need to verify that `newData` is never a falsy but valid value (and know which values are falsy).

christinaroise: In a different case, I'd argue the same, but my reason for wanting to use `||` in this specific context is mainly what outcome we want:

```
const data = newData || initialSalaryAndBenefits.data;
// If newData is "", it will fall back to initialSalaryAndBenefits.data
```

```
const data = newData ?? initialSalaryAndBenefits.data;
// If newData is "", it will be used as is, since it's not null or undefined
```

Because we know that the  SalaryAndBenefits component we're returning is expecting to receive either a valid object or null, we want to make sure that it never receives a potential `""` or `0`. 

So by using `||`, we guarantee that any falsy value—other than `null` or `undefined`—will trigger the fallback, helping us catch issues early if the data model changes unexpectedly.

mathiazom: I'm not sure I follow the last point. With `||`, an unexpected `""` would silently fail and return the fallback. That would prevent the component from crashing, but it would not be clear why the component is not receiving any new data. I think an unexpected data format change, which violates a type assumption, should trigger a clear error.

(`||` would also use fallback for all falsy values, including `null` and `undefined`)

christinaroise: Ok so I did some thinking and digging. I think we both have pretty valuable concerns here. 

I don't want an error to happen in production, but we also want an error to be caught in dev. I propose we make sure that `initialSalaryAndBenefits.data` is **always** the fallback in **production**, but we can add a safecheck in dev:

```
if (process.env.NODE_ENV !== 'production') {
  if (typeof newData !== 'object' || newData === null) {
    console.error(`newData has an unexpected type: ${typeof newData}. Expected an object.`);
  }
}
const data = newData || initialSalaryAndBenefits.data;
```

mathiazom: That seems reasonable :+1:, will fix

christinaroise: Maybe in a different PR we could introduce:

```
if (process.env.NODE_ENV !== 'production') {
  if (typeof newData !== 'object' || newData === null) {
    console.error(`newData has an unexpected type: ${typeof newData}. Expected an object.`);
  }
}
```

as universal as it would be used for all Preview components? 🤔

PR #495, opened 2024-08-21
v3 - benefit type field

mathiazom:
Added field for type of benefit. "Basic" option for benefits not fitting any of the other types.
image

List item preview includes benefit type
image

### DIFF
--- a/studio/schemas/documents/benefit.ts
+++ b/studio/schemas/documents/benefit.ts
@@ -4,9 +4,9 @@ import { richText, title } from "../fields/text";
 export const benefitId = "benefit";
 export const benefitTypeId = "benefitType";
 
-const BENEFIT_TYPE_NONE_VALUE = "none";
+const BENEFIT_TYPE_BASIC_VALUE = "basic";
 const BENEFIT_TYPES = [
-  { title: "None", value: BENEFIT_TYPE_NONE_VALUE },
+  { title: "Basic", value: BENEFIT_TYPE_BASIC_VALUE },
   { title: "Gadgets", value: "gadgets" },
   { title: "Bonus", value: "bonus" },
   { title: "Pension", value: "pension" },
@@ -18,12 +18,12 @@ const benefitType = defineField({
   type: "string",
   title: "Benefit Type",
   description:
-    "Choose the type of benefit (use None for generic benefits). Some benefit types include visual graphs that will be displayed together with the text.",
+    "Choose the type of benefit. Some benefit types include visual graphs that will be displayed together with the text.",
   options: {
     list: BENEFIT_TYPES,
     layout: BENEFIT_TYPES.length > 5 ? "dropdown" : "radio",
   },
-  initialValue: BENEFIT_TYPE_NONE_VALUE,
+  initialValue: BENEFIT_TYPE_BASIC_VALUE,
   validation: (Rule) => Rule.required(),
 });
 
--- a/studio/schemas/documents/benefit.ts
+++ b/studio/schemas/documents/benefit.ts
@@ -4,7 +4,9 @@ import { richText, title } from "../fields/text";
 export const benefitId = "benefit";
 export const benefitTypeId = "benefitType";
 
+const BENEFIT_TYPE_NONE_VALUE = "none";
 const BENEFIT_TYPES = [
+  { title: "None", value: BENEFIT_TYPE_NONE_VALUE },
   { title: "Gadgets", value: "gadgets" },
   { title: "Bonus", value: "bonus" },
   { title: "Pension", value: "pension" },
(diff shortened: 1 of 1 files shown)


studio/schemas/documents/benefit.ts
```diff
@@ -1,27 +1,45 @@
-import { defineType } from 'sanity';
-import { richText, title } from '../fields/text';
+import { defineField, defineType } from "sanity";
+import { richText, title } from "../fields/text";
 
 export const benefitId = "benefit";
+export const benefitTypeId = "benefitType";
+
+const BENEFIT_TYPES = [
+  { title: "Gadgets", value: "gadgets" },
```
christinaroise, line 8: Let's also add:
  { title: "None", value: "none" },
  
  So that we can allow users to have benefit sections without any visuals like graphs

studio/schemas/documents/benefit.ts
```diff
@@ -1,27 +1,45 @@
...
+  { title: "Bonus", value: "bonus" },
+  { title: "Pension", value: "pension" },
+  { title: "Salary Growth", value: "salaryGrowth" },
+];
+
+const benefitType = defineField({
```
christinaroise, line 14: let's set to required and add an initialValue of: `{ title: "None", value: "none" }`

studio/schemas/documents/benefit.ts
```diff
@@ -1,27 +1,45 @@
...
     select: {
       title: title.name,
+      type: benefitType.name,
     },
-    prepare({ title }) {
+    prepare({ title, type }) {
       return {
         title,
-        subtitle: "Benefit",
+        subtitle: `Benefit${type ? ` / ${BENEFIT_TYPES.find((o) => o.value === type)?.title ?? "Unknown benefit type"}` : ""}`,
```
christinaroise, line 39: Ok, so this might be changed with the initialValue and required ruleset of BenefitType.. but I'd also like to simplify this for easier understanding of what's happening. 

```
const benefitType = BENEFIT_TYPES.find((o) => o.value === type)?.title;
const subtitle = type ? benefitType :  "Unknown benefit type";  // though I'm not sure this is needed now that BenefitType has initialValue
```

```suggestion
        subtitle: subtitle,
```

studio/schemas/documents/benefit.ts
mathiazom: The user already has the option to leave the field blank, so that the document does not have the benefitType property. But an explicit option is probably a good idea :+1:

christinaroise: Yea, I think that gives a better UX as it has an actual "none" value

studio/schemas/documents/benefit.ts
mathiazom: "Unknown benefit type" was mostly intended for catching strings not matching any of the current benefit types (which could change over time).

Probably a good idea to split it up :+1: but the suggested change is missing the "Benefit / " prefix. Do you not want that?

studio/schemas/documents/benefit.ts
mathiazom: Is "None" the right word though? What about "Generic" or "Other"?

christinaroise: Yes, I believe so. The user don't want the benefit to have a type so it chooses "None"

studio/schemas/documents/benefit.ts
christinaroise: Hm, no we already know it'll be added to the list of benefits, so it's kind of redundant

mathiazom: Ok, but the subtitle should probably not be "None" for benefits with type "None"? Maybe not have a subtitle for those?

christinaroise: We could do "Type of benefit: Gadgets" as subtitle but that would also set it to "Type of Benefit: none" ... 🤔

mathiazom: That was the reason for having "Benefit" for "None" and "Benefit / {type}" for the others

christinaroise: Hm what about "Basic" if it's none.. 🤔 perhaps we could then just use "Basic" as the value as well.

mathiazom: That works :+1: can always adjust the wording later

studio/schemas/documents/benefit.ts
```diff
@@ -1,27 +1,52 @@
...
+  name: benefitTypeId,
+  type: "string",
+  title: "Benefit Type",
+  description:
+    "Choose the type of benefit. Some benefit types include visual graphs that will be displayed together with the text.",
+  options: {
+    list: BENEFIT_TYPES,
+    layout: BENEFIT_TYPES.length > 5 ? "dropdown" : "radio",
+  },
+  initialValue: BENEFIT_TYPE_BASIC_VALUE,
```
christinaroise, line 26: is it correct that initial value can be a string while the options are object values?

mathiazom: Yes, `initialValue` must be a `string` (or a function resolving to a `string`)

christinaroise (approved):

PR #496, opened 2024-08-21
v3 - schema option to toggle salary calculator

mathiazom:
image

### DIFF
--- a/studio/schemas/deskStructure.ts
+++ b/studio/schemas/deskStructure.ts
@@ -14,7 +14,7 @@ import { soMeLinksID } from "./documents/socialMediaProfiles";
 import { siteSettingsID } from "./documents/siteSettings";
 import { postId } from "./documents/post";
 import { legalDocumentID } from "./documents/legalDocuments";
-import { salaryAndBenefitsId } from './documents/salaryAndBenefits';
+import { salaryAndBenefitsId } from "./documents/salaryAndBenefits";
 
 export default (S: StructureBuilder) =>
   S.list()
@@ -27,7 +27,7 @@ export default (S: StructureBuilder) =>
           S.document()
             .schemaType(siteSettingsID)
             .documentId(siteSettingsID)
-            .title("Site Settings")
+            .title("Site Settings"),
         ),
       S.listItem()
         .title("Legal Documents")
@@ -44,7 +44,7 @@ export default (S: StructureBuilder) =>
           S.document()
             .schemaType("navigationManager")
             .documentId("navigationManager")
-            .title("Navigation Manager")
+            .title("Navigation Manager"),
         ),
       S.listItem()
         .title("Pages")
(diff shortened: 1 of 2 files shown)


studio/schemas/deskStructure.ts
```diff
@@ -14,7 +14,7 @@ import { soMeLinksID } from "./documents/socialMediaProfiles";
 import { siteSettingsID } from "./documents/siteSettings";
```
christinaroise, line 14: ok, we need to add common linting! 💡

studio/schemas/deskStructure.ts
```diff
@@ -64,21 +64,21 @@ export default (S: StructureBuilder) =>
...
               S.listItem()
                 .title("Posts")
                 .icon(ComposeIcon)
                 .child(S.documentTypeList(postId).title("Posts")),
-            ])
+            ]),
         ),
       S.listItem()
         .title("Salary and Benefits")
         .icon(HeartIcon)
```
christinaroise, line 77: [this is not an actual PR comment] would it be nice with a ConfettiIcon instead of a heart? We could also ask design, this is as much part of the product as the actual website. 🤔  Perhaps we should let them in on this later on.

studio/schemas/documents/salaryAndBenefits.ts
```diff
@@ -1,27 +1,37 @@
...
+import { defineField, defineType } from "sanity";
+import { pageSlug } from "../schemaTypes/slug";
+import seo from "../objects/seo";
+import { title } from "../fields/text";
+import { benefitId } from "./benefit";
 
-export const salaryAndBenefitsId = "salaryAndBenefits";
+export const salaryAndBenefitsId = {
+  _root: "salaryAndBenefits",
+  showSalaryCalculator: "showSalaryCalculator",
```
christinaroise, line 9: since we don't know if this is needed just yet, maybe we can add it later? 

it is not wrong in any sense, and actually a good approach, but then I would also want to make the same adjustments to other id variables.. 

I also never investigated why using const in groq returned an error. the thought was to use them in queries, but it sort of went forgotten. 

We can look into it and then whatever solution works we can apply this to the other queries as well?

Aaaand with the new versions we could also adjust with defineQuery 🤩 .. then again a bigger adjustment maybe: 
https://www.sanity.io/changelog/53b04484-04b7-4af8-a796-2ba8a819bb00#f6f340aa0152

christinaroise (approved):

studio/schemas/deskStructure.ts
mathiazom: This is actually me fixing my own formatting from earlier, so I think we have the same now (using Prettier support in WebStorm)

studio/schemas/deskStructure.ts
christinaroise: We can invite the designers and let them make decisions here!

PR #499, opened 2024-08-21
v3 - common util for preview page data type checking

mathiazom:
Ref: https://github.com/varianter/variant.no/pull/492/files/4119c839acb7788e9a888b30429f7b8de9e9e745..a660cb15ddc80baf1245409f37e8140fa1af1523#r1724852364

### DIFF
--- a/src/utils/draftmode.ts
+++ b/src/utils/draftmode.ts
@@ -11,13 +11,3 @@ export function getDraftModeInfo(): DraftModeInfo {
   const perspective = isDraftMode ? "previewDrafts" : "published";
   return { isDraftMode, perspective };
 }
-
-export function checkDraftDataIfDevelopment(data: unknown) {
-  if (process.env.NODE_ENV === "development") {
-    if (typeof data !== "object" || data === null) {
-      console.error(
-        `Draft data has an unexpected type: ${typeof data}. Expected an object.`,
-      );
-    }
-  }
-}
--- a/src/utils/draftmode.ts
+++ b/src/utils/draftmode.ts
@@ -11,3 +11,13 @@ export function getDraftModeInfo(): DraftModeInfo {
   const perspective = isDraftMode ? "previewDrafts" : "published";
   return { isDraftMode, perspective };
 }
+
+export function checkDraftDataIfDevelopment(data: unknown) {
+  if (process.env.NODE_ENV === "development") {
+    if (typeof data !== "object" || data === null) {
+      console.error(
+        `Draft data has an unexpected type: ${typeof data}. Expected an object.`,
+      );
+    }
+  }
+}
--- a/src/blog/BlogPreview.tsx
+++ b/src/blog/BlogPreview.tsx
@@ -5,7 +5,7 @@ import { BLOG_PAGE_QUERY } from "studio/lib/queries/pages";
 import { Suspense } from "react";
 import { Blog } from "./Blog";
 
-import { checkPreviewDataIfDevelopment } from "../utils/preview";
+import { validateDraftDataInDevelopment } from "../utils/preview";
 
 interface BlogPreviewProps {
   initialBlog: QueryResponseInitial<BlogPage>;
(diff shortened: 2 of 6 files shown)


src/utils/draftmode.ts
```diff
@@ -11,3 +11,13 @@ export function getDraftModeInfo(): DraftModeInfo {
   const perspective = isDraftMode ? "previewDrafts" : "published";
   return { isDraftMode, perspective };
 }
+
+export function checkDraftDataIfDevelopment(data: unknown) {
```
christinaroise: since we're validating the data how about giving it a more specific name? 

```suggestion
export function validateDraftDataInDevelopment(data: unknown): void {
```

I'd also add void the explicit return type, just to clarify that the function is not expected to return any value.

christinaroise (commented):
should we include an update with the usage of the validation check for all preview components, or add them in a separate PR? 

in this PR you're only including blog, postPage and SalaryAndBenefits, but most components have Previews.. 🤔

christinaroise:
Aah you mean for sections?

```
  const articleSection = newData
    ? (newData.sections.find(
        (section, index) =>
          section._type === "article" && index === sectionIndex
      ) as ArticleSection)
    : (initialData.data.sections.find(
        (section, index) =>
          section._type === "article" && index === sectionIndex
      ) as ArticleSection);
```

these could actually be made cleaner with say:

```
  const newSection = newData?.sections.find(
        (section, index) =>
          section._type === "article" && index === sectionIndex
      ) as ArticleSection
  const initialSection = initialData.data.sections.find(
    (section, index) => section._type === "article" && index === sectionIndex
  ) as ArticleSection;

validateDraftDataInDevelopment(newSection);
  
  const articleSection = newSection || initialSection;
```

Still debating back and forth on whether we want to keep it the same PR here or not 🤔

mathiazom:
> Aah you mean for sections?


Was thinking about e.g. `LegalPreview`, `FooterPreview` and `HeaderPreview`

christinaroise:
Aah, yea I think those are legacy 💡  and shouldn't have a better handling. wdyt?

mathiazom:
> Aah, yea I think those are legacy 💡 and shouldn't have a better handling. wdyt?

Then we can probably leave them, at least for this PR. 

For `ArticlePreview`, how about this?

```
  validateDraftDataInDevelopment(newData);
  const pageData = newData || initialData.data;

  const articleSection = pageData.sections.find(
    (section, index) => section._type === "article" && index === sectionIndex,
  ) as ArticleSection;
```

christinaroise:
Ops, meant to write 

> Aah, yea I think those are legacy 💡 and **should** have a better handling. wdyt?


Anyway, love this:
```

  validateDraftDataInDevelopment(newData);
  const pageData = newData || initialData.data;

  const articleSection = pageData.sections.find(
    (section, index) => section._type === "article" && index === sectionIndex,
  ) as ArticleSection;
```

mathiazom:
Looking at this again, doesn't the `initial` prop of useQuery already cover initial data? Or is the type of `data` still nullable?

mathiazom:
> > Looking at this again, doesn't the initial prop of useQuery already cover initial data? Or is the type of data still nullable?
> 
> Can you walk me through your question here?

Since we are already defining the initial data for `useQuery` via the `initial` prop, can we just use 
```js
const pageData = newData
```
instead of
```js
const pageData = newData || initialData.data
``` 
It seems to work for `SalaryAndBenefitsPreview.tsx` at least 

```jsx
const SalaryAndBenefitsPreview = ({
  initialSalaryAndBenefits,
}: SalaryAndBenefitsPreviewProps) => {
  const { data } = useQuery<SalaryAndBenefitsPayload>(
    SALARY_AND_BENEFITS_PAGE_QUERY,
    { slug: initialSalaryAndBenefits.data.slug.current },
    { initial: initialSalaryAndBenefits },
  );
  
  return (
    <Suspense>
      <SalaryAndBenefits salaryAndBenefits={data} />
    </Suspense>
  );
};
```

christinaroise:
Hm good catch! 💡 

I wonder why the preview example from sanity doesn't just use newData 🤔  Could it be because  in case the useQuery fails for whatever reason? But then that should give a technical error.

christinaroise (approved):

Issue #498, opened 2024-08-21
Validate slug characters


mathiazom:
Slug value specified by content editors should be valid in terms of url paths. This should be explained and validated in the Sanity Studio form

mathiazom:
Closed by #500

PR #500, opened 2024-08-21
v3 - validate page slug chars

mathiazom:
#498

Making sure to convert Norwegian chars to latin equivalents in slug generation + remove URL-unsafe chars

Also stronger validation of URL-unsafe chars for non-generated slugs

### DIFF
--- a/studio/schemas/schemaTypes/slug.ts
+++ b/studio/schemas/schemaTypes/slug.ts
@@ -55,8 +55,6 @@ function createSlugField(source: string) {
 }
 
 const pageSlug = createSlugField("page");
-const titleSlug = createSlugField("basicTitle");
-const blogSlug = titleSlug;
-const legalSlug = titleSlug;
+const pageTitleSlug = createSlugField("basicTitle");
 
-export { pageSlug, titleSlug, blogSlug, legalSlug };
+export { pageSlug, pageTitleSlug };
--- a/studio/schemas/schemaTypes/slug.ts
+++ b/studio/schemas/schemaTypes/slug.ts
@@ -55,6 +55,6 @@ function createSlugField(source: string) {
 }
 
 const pageSlug = createSlugField("page");
-const pageTitleSlug = createSlugField("basicTitle");
+const titleSlug = createSlugField("basicTitle");
 
-export { pageSlug, pageTitleSlug };
+export { pageSlug, titleSlug };
--- a/studio/schemas/schemaTypes/slug.ts
+++ b/studio/schemas/schemaTypes/slug.ts
@@ -2,7 +2,7 @@ import { defineField, SlugValidationContext } from "sanity";
 
 async function isSlugUniqueAcrossAllDocuments(
   slug: string,
-  context: SlugValidationContext
+  context: SlugValidationContext,
 ) {
   const { document, getClient } = context;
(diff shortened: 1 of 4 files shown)


studio/schemas/schemaTypes/slug.ts
```diff
@@ -19,26 +19,44 @@ async function isSlugUniqueAcrossAllDocuments(
...
 
 const pageSlug = createSlugField("page");
-const blogSlug = createSlugField("basicTitle");
-const legalSlug = createSlugField("basicTitle");
+const titleSlug = createSlugField("basicTitle");
+const blogSlug = titleSlug;
+const legalSlug = titleSlug;
 
-export { pageSlug, blogSlug, legalSlug };
+export { pageSlug, titleSlug, blogSlug, legalSlug };
```
christinaroise, line 62:
```suggestion
export { pageSlug, pageTitleSlug };
```
then you can replace blogSlug, legalSlug with titleSlug

studio/schemas/schemaTypes/slug.ts
```diff
@@ -19,26 +19,42 @@ async function isSlugUniqueAcrossAllDocuments(
...
   });
 }
 
 const pageSlug = createSlugField("page");
-const blogSlug = createSlugField("basicTitle");
-const legalSlug = createSlugField("basicTitle");
+const pageTitleSlug = createSlugField("basicTitle");
 
-export { pageSlug, blogSlug, legalSlug };
+export { pageSlug, pageTitleSlug };
```
christinaroise, line 60: just be second guessing myself here.

do you agree or disagree that this makes sense?

my thought was that if it's pagetitle, it explicitly says it's the slug generated from the page title and to be used within page schemas, and if it's pageSlug it's to be used in pages without a title, but a page name because these are names used for sanity studio identification and for slug as it can be different from the title used on screen.. 

which again makes me think that we shouldn't be using the title in salaryAndBenefits to generate the slug as we want the slug to be `/kalkulator` but the title might be "Hos variant har vi drit kule goder" 💡  Ofc the user can manually set the slug by not generating from title.. 🤔

christinaroise: for page we needed to generate because title is set in section.hero.. 💡 so perhaps it's ok that we're using title anyway in salary and benefits!

christinaroise: But perhaps the value should then be 

`titleSlug` and not `pageTitleSlug` like original?

mathiazom: I agree that `titleSlug` is better (or even `basicTitleSlug`). I just kept the other exports to avoid changing other parts of the code, but it retrospect that might just make it more confusing. 

We don't have anything other than title to base the slug on, but we could always make the slug have some static value as source for "Generate". I think the title based slug is fine though.

christinaroise: Yea, I think titleSlug or basicTitleSlug might be better, yes!

mathiazom: Are we good then?

christinaroise (approved):

studio/schemas/schemaTypes/slug.ts
christinaroise: Yes! I forgot to set "approved"

PR #506, opened 2024-08-22
Feature/salary and benefits calculator

anemne:
## Short Description

Added simple calculation method for calculator and added it in benefits and salary. Calculator will hide if showSalaryCalculator is set to false. 
---

If applicable, please include screenshots or a short video showcasing the changes you have made.

*Insert images or videos here.*
image

---

## Checklist

- [ ] **Documentation**: Relevant documentation has been added or updated (if applicable).
- [ ] **Testing**: Have you tested your changes thoroughly?
  - Please list the types of tests you've run (unit, integration, manual, etc.):

---

## Additional Notes

We still need to do some error handling if selected examination year 2024 or more. Also, the existing payscale data is a bit off.


src/salaryAndBenefits/utils/salaryData.ts
mathiazom, line 1: Why are the salaries given as strings and not numbers?

src/salaryAndBenefits/components/SalaryCalculator/SalaryCalculator.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const options: IOption[] = [
+  { id: "bachelor", label: "Bachelor", currentChecked: false, disabled: false },
+  { id: "master", label: "Master", currentChecked: true, disabled: false },
+];
+
+export default function SalaryCalculator() {
+  const thisYear = new Date().getFullYear();
+  const [examinationYear, setExaminationYear] = useState(thisYear);
+  const [selectedOption, setSelectedOption] = useState("master");
```
mathiazom, line 25: Should give this a more specific name like `selectedDegree`

src/salaryAndBenefits/utils/calculateSalary.ts
```diff
@@ -0,0 +1,30 @@
...
+    degree: string
+  ): string {
+
+    if (examinationYear >= currentYear) {
+        //TODO: add error handling for data that is not in the payscale
+        console.log("Examination year cannot be in the future.");
+    }      
+
+    const adjustedYear = (examinationYear + (degree === "Bachelor" ? 1 : 0)).toString();
+    return salaryPayscale[currentYear][adjustedYear];
```
mathiazom, line 23: Shouldn't `currentYear` also be converted to a string?

mathiazom: JavaScript handles indexes as string internally anyway, so it will run regarless 🙃

src/salaryAndBenefits/SalaryAndBenefits.tsx
```diff
@@ -1,16 +1,72 @@
...
+  const [examinationYear, setExaminationYear] = useState(thisYear);
+  const [selectedOption, setSelectedOption] = useState("master");
+  const [salary, setSalary] = useState(
+    calculateSalary(thisYear, examinationYear, selectedOption)
+  );
+
+  const handleChange = (name: string, value: string) => {
+    if (name === "examinationYear") {
+      setExaminationYear(parseInt(value));
+      setSalary(calculateSalary(thisYear, parseInt(value), selectedOption));
```
mathiazom, line 40: Could consider a `useMemo` for salary (or even just `const salary = calculateSalary(...)` instead of `useState` and `setSalary`

src/salaryAndBenefits/SalaryAndBenefits.tsx
```diff
@@ -1,16 +1,72 @@
...
       <Text type="h1">{salaryAndBenefits.basicTitle}</Text>
+      {salaryAndBenefits.showSalaryCalculator && 
+      <div className={styles.calculator}>  
+      <RadioButtonGroup
+        id="degree-group"
+        label="Choose your degree"
+        options={options}
+        onValueChange={handleOptionChange}
+      />
+      <InputField
```
mathiazom, line 60: `InputField` should have `type="number"`

mathiazom: and ideally a `min` and `max`, but that requires changes to `InputField`

example from outdated PR for year input:
https://github.com/varianter/variant.no/blob/1cdd430ecdf9500e995ee2c426ea870d80d61dd6/src/components/forms/inputYear/InputYear.tsx#L31

src/salaryAndBenefits/utils/salaryData.ts
anemne: good question! Just converted from old solution. The numbers are also a bit off (example: 2020 have calculations for 2022.. etc.). Should we store data from only 2024 for now?

src/salaryAndBenefits/utils/calculateSalary.ts
anemne: Does typescript do this as well?

src/salaryAndBenefits/SalaryAndBenefits.tsx
anemne: I agreethat usememo is a better solution! Think we need this as degree and examinationyear could change

src/salaryAndBenefits/utils/salaryData.ts
mathiazom: 2024 is enough for now, we can always add more later

src/salaryAndBenefits/utils/calculateSalary.ts
mathiazom: No difference at runtime, so yes. But in theory some checks might complain about the type difference.

src/salaryAndBenefits/SalaryAndBenefits.tsx
anemne: We need to figure out if we should use inputfield, dropdown or other. We need to use whats bes regarding UU

src/salaryAndBenefits/utils/salaryData.ts
```diff
@@ -0,0 +1,41 @@
+export const payscale = {
+    '2024': {
+      "2024" : "600000",
```
mathiazom, line 3: Still think the salary values should be numbers

src/salaryAndBenefits/SalaryAndBenefits.tsx
```diff
@@ -1,16 +1,35 @@
...
+import SalaryCalculator from "./components/salaryCalculator/SalaryCalculator";
+import { useState } from "react";
+import { calculatePension } from "./utils/calculateSalary";
 
 interface SalaryAndBenefitsProps {
   salaryAndBenefits: SalaryAndBenefitsPayload
 }
 
 const SalaryAndBenefits = ({salaryAndBenefits}: SalaryAndBenefitsProps) => {
+  const [salary, setSalary] = useState<string | null>(null);
```
mathiazom, line 16: This should also be passed around as a number

christinaroise: Hm, so normally I'd like bs logic related to the calc. to be within the component itself, but because we know that the result from the calculator will affect graphs and other visuals within the page comp, I believe it to be best to keep it all on parent level instead of having some here and some there which are also dependant on one another. 

So let's keep `SalaryCalculator` as a pure UI component. 

global states could be a possible solution, but we don't yet know the full scope of the website, so it's nice to keep it as simple as possible for now.

src/salaryAndBenefits/components/salaryCalculator/SalaryCalculator.tsx
```diff
@@ -0,0 +1,68 @@
...
+import InputField from "src/components/forms/inputField/InputField";
+import { RadioButtonGroup } from "src/components/forms/radioButtonGroup/RadioButtonGroup";
+import { calculateSalary } from "src/salaryAndBenefits/utils/calculateSalary";
+import Button from "src/components/buttons/Button";
+
+interface IOption {
+  id: string;
+  label: string;
+  disabled?: boolean;
+  currentChecked: boolean;
```
christinaroise, line 14:
```suggestion
  currentSelected: boolean;
```

src/salaryAndBenefits/components/salaryCalculator/SalaryCalculator.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+const options: IOption[] = [
+  { id: "bachelor", label: "Bachelor", currentChecked: false, disabled: false },
```
christinaroise, line 18:
```suggestion
  { id: "bachelor", label: "Bachelor", currentSelected: false, disabled: false },
```

christinaroise: Linguist in me: checked is for checkbox while radio buttons are selected 💡

src/salaryAndBenefits/components/salaryCalculator/SalaryCalculator.tsx
```diff
@@ -0,0 +1,68 @@
...
+];
+
+interface SalaryCalculatorProps {
+    onSalaryUpdate: (salary: string) => void;
+  }
+
+export default function SalaryCalculator({ onSalaryUpdate }: SalaryCalculatorProps) {
+  const thisYear = new Date().getFullYear();
+  const [examinationYear, setExaminationYear] = useState(thisYear);
+  const [selectedDegree, setSelectedDegree] = useState("master");
```
christinaroise, line 29: Side comment: Why do we default to master? I think we should have a chat with Even perhaps in terms of inclusivity when we default to a higher level education.

christinaroise, line 29: You can combine to one useState like so:

```
interface FormState {
  examinationYear: number
  selectedDegree: "bachelor" | "master", // btw we should really revisit this in terms of inclusivity....
}

/////

const [formState, setFormState] = useState<FormState>({
  examinationYear: thisYear,
  selectedDegree: "master",
});

// To update examinationYear
const updateExaminationYear = (newYear) => {
  setFormState((prevState) => ({
    ...prevState,
    examinationYear: newYear,
  }));
};

// To update selectedDegree
const updateSelectedDegree = (newDegree) => {
  setFormState((prevState) => ({
    ...prevState,
    selectedDegree: newDegree,
  }));
};
```

src/salaryAndBenefits/components/salaryCalculator/SalaryCalculator.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <InputField
+        label="year"
+        name="examinationYear"
+        type="number"
+        value={examinationYear.toString()}
+        onChange={handleChange}
+        required
+      />
+      <Button type="secondary" size="small">
+        submit
```
christinaroise, line 64:
```suggestion
        Submit
```

src/salaryAndBenefits/utils/salaryData.ts
christinaroise: It depends on what we add once a year I believe, we might want to adjust the script.

src/salaryAndBenefits/utils/calculateSalary.ts
```diff
@@ -0,0 +1,29 @@
...
+    };
+  }
+  
+  const salaryPayscale: PayScale = payscale;
+  
+  export function calculateSalary(
+    currentYear: number,
+    examinationYear: number,
+    degree: string
+  ): string {
```
mathiazom, line 15: Should return a number

src/salaryAndBenefits/utils/calculateSalary.ts
```diff
@@ -0,0 +1,29 @@
...
+        //TODO: add error handling for data that is not in the payscale
+        console.log("Examination year cannot be in the future.");
+    }      
+
+    const adjustedYear = (examinationYear + (degree === "Bachelor" ? 1 : 0));
+    return salaryPayscale[currentYear.toString()][adjustedYear];
+   
+  }
+
+  export function calculatePension(salary: string): string {
```
mathiazom, line 27: Should return a number 🙃

src/salaryAndBenefits/utils/salaryData.ts
christinaroise: I'd argue against this. I'd want to keep historic data as they might be needed in other or future calculations.

src/salaryAndBenefits/utils/calculateSalary.ts
```diff
@@ -0,0 +1,29 @@
...
+    examinationYear: number,
+    degree: string
+  ): string {
+
+    if (examinationYear >= currentYear) {
+        //TODO: add error handling for data that is not in the payscale
+        console.log("Examination year cannot be in the future.");
+    }      
+
+    const adjustedYear = (examinationYear + (degree === "Bachelor" ? 1 : 0));
```
christinaroise, line 22: hmm could we simplify this for readability? 
```suggestion
const constValue = degree === "Bachelor" ? 1 : 0. // give this a proper name
    const adjustedYear = (examinationYear + constValue;
```

src/salaryAndBenefits/components/salaryCalculator/SalaryCalculator.tsx
anemne: Is it better to set it to bachelor maybe?

mathiazom: [Legacy page](https://www.variant.no/kalkulator) defaults to bachelor 🤓

src/salaryAndBenefits/utils/calculateSalary.ts
mathiazom: Not convinced this is more readable, at least not without a very good name

src/salaryAndBenefits/components/salaryCalculator/SalaryCalculator.tsx
```diff
@@ -0,0 +1,68 @@
...
+import Button from "src/components/buttons/Button";
+
+export type Degree = "bachelor" | "master";
+
+const degreeOptions: IOption[] = [
+  {
+    id: "bachelor",
+    label: "Bachelor",
+    currentSelected: true,
+    disabled: false,
```
mathiazom, line 16: No need to specify `disabled` here, default is `false`

src/components/forms/inputField/InputField.tsx
```diff
@@ -56,6 +60,8 @@ const InputField = ({
         autoComplete={autoComplete}
         autoCorrect={autoCorrect}
         type={type}
+        max={max}
+        min={min}
         className={styles.input}
         spellCheck={spellCheck}
         value={value}
```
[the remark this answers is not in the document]
mathiazom: Should note that `max` and `min` are relevant for more than just `number`, specifically `date`, `month`, `week`, `time`, `datetime-local` and `range`. Since this is a general component, we should not assume the only use case is for numbers.

christinaroise (approved):

Issue #527, opened 2024-08-27
Character-count in Sanity Studio text field


mathiazom:
Some text fields, like for SEO, have validation on the number of total characters. It would be nice to provide the content creator with the actual count to give an idea of how far away they are from the limit.

Example:

https://github.com/user-attachments/assets/29952f36-4d3e-43a6-9986-91018925da69

(source: https://www.youtube.com/watch?v=ubxjQnnpS-4&t=53s)

Example for RichText: https://www.linkedin.com/posts/knutmelvaer_i-wanted-a-small-character-count-below-a-activity-7100204622690070528-rvSl?utm_source=share&utm_medium=member_desktop

mathiazom:
Closed by #529

PR #529, opened 2024-08-27
v3 - custom string input with character count

mathiazom:
Further explanation is found here: #527 

Composes the default string input rendering with a simple character count, with required max count. Added to most of the fields with `type: "string"`, but can also be expanded to other fields later.

Also renamed all instances of the validation function argument `Rule` to `rule` (since this is a variable and not a type or class)

image

image

image

---

## Checklist

  - Please list the types of tests you've run (unit, integration, manual, etc.):

---

## Additional Notes

Other comments relevant to this pull request.


christinaroise (commented):
I know this didn't take too long, but it'd be nice if we followed Monday Comittments more strictly in terms of not picking tasks outside of our goals. 

I also believe there are several other places in the repository with character limit in need of the count.

mathiazom:
> I know this didn't take too long, but it'd be nice if we followed Monday Comittments more strictly in terms of not picking tasks outside of our goals.

I agree, but my other tasks were blocked and this was almost finished from last week.

mathiazom:
> I also believe there are several other places in the repository with character limit in need of the count.

Should these be added in this PR or left for later?

christinaroise:
>I agree, but my other tasks were blocked and this was almost finished from last week.

I noticed after some digging on slack! We need a better system in terms of alerting others when we're blocked. Especially when we're working on tasks towards out goal for the week.

christinaroise:
>Should these be added in this PR or left for later?

I'd do them all in the same PR 👍

christinaroise:
@mathiazom maybe convert to draft until you have the time to add char.limit to the remaining places?

studio/schemas/fields/text.ts
```diff
@@ -53,7 +58,10 @@ export const optionalSubtitle = defineField({
...
   title: "Subtitle",
   type: "string",
-  validation: (Rule) => Rule.max(60),
+  validation: (rule) => rule.max(60),
+  components: {
+    input: (props) => StringInputWithCharacterCount({ ...props, maxCount: 60 }),
+  },
 });
 
 export const richTextID = "richText";
```
anemne, line 67: Will this field also have a character count?  
Screenshot 2024-09-11 at 08 48 08

anemne (approved):

studio/schemas/fields/text.ts
mathiazom: Rich text fields are not supported yet. The calculation is a little more complex since the text is stored as an array of blocks instead of a regular string. I have defined a separate issue for this: #605

studio/components/StringInputWithCharacterCount.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box, Stack, Text } from "@sanity/ui";
```
christinaroise, line 1: before we merge can you make sure it passes wcag 2.1? (plenty of studio is just bleh, but we can give it a try since it's a custom component
https://designsystem.digital.gov/components/character-count/accessibility-tests/

mathiazom: Updated now to show the remaining characters, with a red text when over limit. Also added `aria-describedby` to announce allowed number of characters. `aria-live` has been set on the counter, but I have struggled to get the screen reader to announce it reliably (intercepted by text field description, which might be Sanity's fault).

christinaroise: Yeah, I believe this is on Sanity. Great job though!

Issue #547, opened 2024-08-30
Legge til BenefitsSchema under referansene i Compensations


anemne:
BenefitsSchema skal inn under Compensation

Issue #544, opened 2024-08-30
Definere Compensations under Static Pages: [Compensations(office: Oslo), Compensations(office: Trondheim, Bergen), ...]


anemne:
```[tasklist]
### Tasks
- [x] https://github.com/varianter/variant.no/issues/568
- [ ] https://github.com/varianter/variant.no/issues/567
```

PR #551, opened 2024-08-30
v3 - replace SalaryAndBenefits with Compensations list

mathiazom:
#544, #547

Replaced single `SalaryAndBenefits` with `Compensations` document type allowing multiple instances. This will later be used to allow each office to have their own `Compensations` document.

---

image

---

## Checklist


studio/lib/payloads/compensations.ts
```diff
@@ -8,7 +8,7 @@ export interface Benefit {
   richText: PortableTextBlock[];
```
christinaroise, line 8: to be solved in a diff PR. perhaps it makes sense to cleanup our payload structure. so put static pages in one and dynamic pages in another.. I dunno. We can talk about it later

studio/schemas/documents/salaryAndBenefits.ts
```diff
@@ -1,15 +1,17 @@
 import { defineField, defineType } from "sanity";
```
christinaroise, line 1: since it's deprecated and we haven't launched yet, you can delete.

christinaroise: ⚠️ make sure any content is deleted in PROD first ⚠️

christinaroise (approved):

christinaroise:
I approved, but take a look at comments before merging. 😊

studio/schemas/documents/salaryAndBenefits.ts
mathiazom: This is not just an issue after launch, it also makes it harder to move between versions of the codebase during development. Especially since we don't have any migration scripts. We can still remove the old schemas later when we know all the dependencies on the old schemas have been removed.

christinaroise: As long as you make a note to fix

PR #552, opened 2024-08-30
v3 — add document translations to shared

christinaroise:
This PR is adding [document translations](https://www.sanity.io/docs/localization#4bc811debdde) to our shared studio. 

The way translations for Sanity work is that Sanity creates language specific versions of a document, that would result in:
- a document
- the English version of the document 
- the Norwegian version of the document 
- the Swedish version of the document 

I've structured our custom deskStructure to only show the English version of the document as default, and added the option to translate this document. So that in theory we'll only create: 
- a document with English as the base language
- the Norwegian version of the document 
- the Swedish version of the document

The plan is that frontend will fetch the content in it's defined base language (Norwegian for Variant Norge and Swedish for Variant Sverige) and look for the translation in English. It's unclear whether or not Variant Norge will include Swedish in their translations or not. 

### Demo

To create a costumer case it has to first be created in English
Screenshot 2024-09-10 at 14 48 52

Then you can choose what language to translate the document to
Screenshot 2024-09-10 at 14 50 04

And then you add the translated content (Later we can add an AI plugin for document translations)
Screenshot 2024-09-10 at 14 51 28


studioShared/sanity.config.ts
```diff
@@ -1,16 +1,43 @@
...
+    }),
+    visionTool({ defaultApiVersion: apiVersion }),
+    documentInternationalization({
+      // TODO: a function that takes the client and returns a promise of an array of supported languages
+      // MUST return an "id" and "title" as strings
+      // supportedLanguages: (client) => client.fetch(`*[_type == "language"]{id, title}`),
+      supportedLanguages: i18n.languages,
+      schemaTypes: [customerCaseID],
+      languageField: `language`,
+      metadataFields: [defineField({ name: "slug", type: "slug" })],
```
christinaroise, line 35: unsure of this just yet

studioShared/sanity.config.ts
```diff
@@ -1,16 +1,43 @@
...
+  plugins: [
+    structureTool({
+      structure: deskStructure,
+    }),
+    visionTool({ defaultApiVersion: apiVersion }),
+    documentInternationalization({
+      // TODO: a function that takes the client and returns a promise of an array of supported languages
+      // MUST return an "id" and "title" as strings
+      // supportedLanguages: (client) => client.fetch(`*[_type == "language"]{id, title}`),
+      supportedLanguages: i18n.languages,
```
christinaroise, line 32: languages will be fetched from Variant Norge Studio and Variant Sverige Studio. So we'll get a list of Norwegian, English, Swedish, English and then filter out duplicate and add as supportedLanguages.

studioShared/deskStructure.ts
```diff
@@ -0,0 +1,17 @@
...
+import { customerCaseID } from "./schemas/documents/customerCase";
+
+export const deskStructure: StructureResolver = (S) =>
+  S.list()
+    .title("Content")
+    .items([
+      S.listItem()
+        .title("Customer cases")
+        .child(
+          S.documentTypeList(customerCaseID)
```
mathiazom, line 12: I think the type in `documentTypeList(customerCaseID)` is ignored when using `.filter(...)`, which is why we have to check type in filter as well. So we might want to use a plain `documentList()` to avoid confusion (I at least wondered why you needed the extra `_type == $type`).

languages.ts
```diff
@@ -0,0 +1,12 @@
+const languages = [
+  { id: "en", title: "English", isDefault: true },
```
mathiazom, line 2: What is the difference between `isDefault` and `base`?

studioShared/studioConfig.tsx
```diff
@@ -12,8 +18,32 @@ const config: WorkspaceOptions = {
...
+    }),
+    visionTool({ defaultApiVersion: apiVersion }),
+    documentInternationalization({
+      // TODO: a function that takes the client and returns a promise of an array of supported languages
+      // MUST return an "id" and "title" as strings
+      // supportedLanguages: (client) => client.fetch(`*[_type == "language"]{id, title}`),
+      supportedLanguages: i18n.languages,
+      schemaTypes: [customerCaseID],
+      languageField: `language`,
+      metadataFields: [defineField({ name: "slug", type: "slug" })],
```
mathiazom, line 38: Maybe import `titleSlug` instead of re-defining the field?

languages.ts
christinaroise: None, I'll remove, but all this will also be removed once we get the languages from Variant Norge and Variant Sverige

studioShared/deskStructure.ts
christinaroise: Yes! Good catch

studioShared/studioConfig.tsx
christinaroise: removing as it's not needed and just came with my careless copy paste! 😅

languages.ts
```diff
@@ -0,0 +1,12 @@
+const languages = [
```
christinaroise, line 1: to be deleted with updates in /studio

mathiazom (approved):

anemne (approved):
👌

PR #569, opened 2024-09-03
cleanup compensation and add the start of bonus

christinaroise:
While adding the `compensationDetails` to the existing `compensation` document, I noticed that our initial architecture was a little flawed. In the original structure, which was "sorted" buy location, we assigned individual titles, slugs, SEO fields, and compensation data (such as bonus, salary, etc.) to the location(s) set in the document. But, on our frontend we don't want different page titles, SEO data or slugs. we only want to show/hide content by location it is meant for. (I'm thinking a TAB-solution)

So in other words our initial setup was a little too overly complex, maybe even redundant, and would most likely lead to inefficient data management.

**Solution**
So, to address this, I restructured the compensation schema to reflect a more streamlined approach. Instead of assigning each location its own document structure, I updated the compensation document to have a unified structure with shared fields at the top level and location-specific compensation data grouped logically.

The key changes include:

- deskStructure.ts: The compensation document now uses an object ({}) structure rather than an array ([]).
- compensation.ts: 
-- Shared fields: title, slug, seo, and other core fields like pension are now shared across all locations.
-- Location-Specific Fields: Fields that differ by location, such as bonusesByLocation, benefitsByLocation, and salaryByLocation, are clearly separated and scoped to handle location-based differences effectively. 

‼️ bonusesByLocation is the only one actually added in this PR, I have added comments for future tasks so it's easy to see whats to be added where

‼️ bonusesByLocation is also **ONLY** been added to show the schema structure with location sorting. I have **NOT** taken the schema type of average bonus into account other than setting it to number. Whatever else it needs to be can be updated

‼️ THIS PR IS NOT ADDING ADJUSTMENT TO THE FRONTEND. That needs to be added as well to prevent crashes. But I want to make sure we're agreeing with this approach first. 

‼️ THIS PR DOES INTRODUCE DATA CONFLICTS WITH PRODUCTION AND STAGING. We can write a script or just manually delete data added to staging before we merge. Since we're not actually working with PROD, I don't mind manual work for now.

**Updated Compensation Structure**
The new structure consists of:
	•	title: Main title for the compensation page
	•	seo: SEO metadata shared across all locations
	•	slug: URL slug for the compensation page
	•	pension: Global pension details
	•	bonusesByLocation: Location-specific bonuses
	•	benefitsByLocation: Location-specific benefits
	•	salaryByLocation: Location-specific salary details

 // moved seo to the bottom, I want to check with Linn-Helen, but I think it makes more sense to add SEO at the bottom once all the content is added and it's clear what title, description, and keywords should consist of.

Screenshot 2024-09-04 at 14 21 23

Screenshot 2024-09-04 at 14 21 34

ERROR state for duplicated locations. An idea is to create a custom component for Bonus. Ideally I'd like for Bonus to dynamically show an average amount field for all locations available, and error states for locations missing an average bonus amount
Screenshot 2024-09-04 at 14 21 59

UPDATE for Company Details -> Company Locations
Screenshot 2024-09-04 at 14 25 36


studio/schemas/documents/compensations.ts
```diff
@@ -15,24 +16,29 @@ const compensations = defineType({
...
     }),
+    {
+      name: "showEstimatedSalary",
+      title: "Show Estimated Salary",
+      type: "boolean",
+      description:
+        "Toggle this option to show or hide the estimated salary for the locations you've selected. Disable this if the feature isn't ready or if you prefer not to display salary details at this time.",
+      initialValue: false,
+    },
     defineField({
```
mathiazom, line 34: Why is benefits not placed within compensationDetails?

studio/schemas/documents/compensations.ts
```diff
@@ -15,24 +16,29 @@ const compensations = defineType({
...
-    seo,
+    locations,
     defineField({
-      name: "showSalaryCalculator",
-      title: "Show Salary Calculator",
-      description: "Should the salary calculator be visible on the page?",
-      type: "boolean",
-      initialValue: true,
+      ...compensationDetails,
+      readOnly: ({ parent }) => !parent?.showEstimatedSalary,
```
mathiazom, line 24: Why do we need to make it read only? This forces the content editor to show the estimated salary in the frontend to edit compensation details.

studio/schemas/documents/compensations.ts
christinaroise: Because benefits are text components not related to data like salary and pension

studio/schemas/documents/compensations.ts
christinaroise: Sanity isn't built on the draft principle, that I assume you want the creator to have ownership over. I don't see a situation where that's something Variant would benefit from. 

So we're setting it to read only to indicate that the fields are irrelevant if you're not interested in showing the calculator in production. 

Nothing is added to production until it is published. So if a creator for some reason would like to add content to the compensationDetails while have benefits in production. They can untoggle the show estimated salary, add content, and then toggle it back to false when ready. 

However, that's not a typical CMS type of workflow. What's more common is to create a new document, add all kinds of data and then unpublish v1 and publish v2 once you're done. Sanity does have a bulk document publishing feature in the growth plan for this.

anemne: What about moving "show estimate toggle button" above "compnesation details", and hide this information until the show estimate salary is set to true?

mathiazom: I would go for hiding instead of readonly in that case.

christinaroise: I agree we can move it above! 

I wouldn1t want to hide the content based on a boolean value in this context. (There are other places I'd probably go back and visit in this repository). Because the hidden content might not be easily accessible for users relying on screen readers or other assistive technologies, I'd want them to see it as read only when disabled (assuming sanity has their a11y up to date). That way they have access to the user experience of the schema, and can easily tell whether or not the toggle should be true or false.

studio/schemas/documents/compensations.ts
mathiazom: It says "financial benefits" in the description though 🤔 I think the main issue is that "Compensation Details" is very generic. If the whole section is disabled by "Show estimated salary", maybe we could call it "Salary Details" instead?

studio/schemas/documents/compensations.ts
mathiazom: Would it be clear why the fields are read-only in that case (e.g. "toggle show salary calculator to enable this field")?

studio/schemas/documents/compensations.ts
christinaroise: "Benefits" and "Financial benefits" are similar but have different definitions still. One is broader and has wider range of perks and advantages provided by an employer, which may include both financial and non-financial elements. While the other refers specifically to benefits that have a direct monetary value like salary, bonus, pension. "Salary details" would refer to the base salary information, and wouldn't really cover pension or bonus.

mathiazom: If "salary" does not cover pension or bonus (which makes sense), then why does disabling "Show estimated salary" make the pension and bonus fields read-only?

christinaroise: I added "estimated Salary" more as the title for the feature.

anemne (approved):

studio/schemas/documents/compensations.ts
```diff
@@ -15,74 +12,59 @@ const compensations = defineType({
...
-    defineField({
-      name: "showSalaryCalculator",
-      title: "Show Salary Calculator",
-      description: "Should the salary calculator be visible on the page?",
-      type: "boolean",
-      initialValue: true,
-    }),
+    // add pension here. pension doesn't rely on locations
+    bonusesByLocation,
+    // add salary here
```
christinaroise, line 24: I want to keep these comments so we know what's expected here once we add pension, benefits and salary

studio/schemas/documents/compensations.ts
```diff
@@ -15,74 +12,59 @@ const compensations = defineType({
...
-        subtitle: previewStringFromLocationsMap(
-          locationsMap,
-          LOCATIONS_PREVIEW_CUTOFF,
-        ),
       };
     },
   },
 });
 
-/**
```
christinaroise, line 62: we most likely want to copy paste this into benefits and salary but removing for now.

anemne:
I agree with this new structure! One question: Shouldn't the pension field also have an office location, or should we add it later if necessary?

christinaroise:
>I agree with this new structure! One question: Shouldn't the pension field also have an office location, or should we add it later if necessary?

If I remember, Sweden and Norway both said that pension will be the same for all locations

studio/schemas/objects/compensations/bonusesByLocation.ts
```diff
@@ -0,0 +1,97 @@
...
+  ],
+  validation: (Rule) =>
+    Rule.custom((bonusesByLocation, context) => {
+      const duplicateCheck = checkForDuplicateLocations(
+        bonusesByLocation as BonusEntry[] | undefined,
+      );
+
+      if (duplicateCheck !== true) return duplicateCheck;
+
+      return true;
```
mathiazom, line 59:
```suggestion
      return duplicateCheck;
```

mathiazom (approved):

Issue #573, opened 2024-09-04
OpenGraph image fallback for each page


mathiazom:
Each page should have an image shown when shared via url. This is achieved via the [opengraph](https://ogp.me) image tag. Currently this is achieved by setting the SEO image field. A fallback can be defined in Company Info, but this is just a generic image. A fallback should be generated based on the page title and description.

mathiazom:
Closed by #574

PR #574, opened 2024-09-04
v3 - OpenGraph fallback image

mathiazom:
Further explanation is found here: #573 

A fallback image is generated for pages without an SEO image specified in Sanity.

This is done through an API route which returns an image response based on given title and description parameters.

Example:

`title`: "Raus, åpen og læreglad"
`description`: "Tjenesteutvikling handler ikke om designskisser, kodelinjer eller infrastruktur, men om samarbeid og forståelse. Åpenhet, ærlighet og endringsvilje."

The following `meta` tag is generated and included in the page `<head/>`:

`<meta property="og:image" content="http://localhost:3000/api/openGraphImage?title=Raus%2C+%C3%A5pen+og+l%C3%A6reglad&amp;description=Tjenesteutvikling+handler+ikke+om+designskisser%2C+kodelinjer+eller+infrastruktur%2C+men+om+samarbeid+og+forst%C3%A5else.+%C3%85penhet%2C+%C3%A6rlighet+og+endringsvilje.%0A">`

With the following image response:

image

---

## Checklist


christinaroise:
Nice approach 👍 

I am picky about this, and I know it's frustrating because we don't have design in place yet. But I'd prefer not to add too much for now as we know most will be replaced with the new brand. Like fonts and colors. 

I would also prefer, if OpenGraph admits it, to refrain from inline styling. 

And I think we can take a look at the folder structure. The app folder should only really contain routes, layouts and pages. So what about a folder for seo related components and utils? Perhaps the utils for seo could also be placed there as well? 


Screenshot 2024-09-05 at 12 37 56

christinaroise:
Good point, what about adding some documentation instead? I think that's valuable information to find in readme

mathiazom:
> Good point, what about adding some documentation instead? I think that's valuable information to find in readme

Are you thinking a basic version without legacy assets, with documentation on how to add that later - or with legacy assets and documentation on how to replace it? I'm leaning more towards the latter. We could inline example code in the README, but I'm concerned it might bloat the documentation.

christinaroise:
This is a nice link to add to docs:
https://developers.google.com/search/docs/appearance/google-images

mathiazom:
> This is a nice link to add to docs: https://developers.google.com/search/docs/appearance/google-images

I don't see it mention OpenGraph specifically, so maybe include this under a more generic "Images best practices" header?

Will add a reference to https://ogp.me under the OpenGraph header.

src/app/api/openGraphImage/OpenGraphImage.tsx
```diff
@@ -0,0 +1,51 @@
+interface OpenGraphImageProps {
```
christinaroise, line 1: I'd like to avoid inline styling, but let's just add a comment saying this is just an example for now i.e mock, and we'll just update later. I'm pretty sure the image will be an illustration and not contain any text or styling.

christinaroise (approved):

src/app/api/openGraphImage/OpenGraphImage.tsx
mathiazom: The CSS support is limited for this element (see https://github.com/vercel/satori?tab=readme-ov-file#css), so I am not sure it supports CSS modules.

Also, the whole point of generating the image is that we can use dynamic data from the page (like title and description). Otherwise we could have just used a static image asset.

Issue #578, opened 2024-09-04
Automatic sitemap


mathiazom:

mathiazom:
https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap#generating-a-sitemap-using-code-js-ts

mathiazom:
https://www.sanity.io/headless-seo/seo-with-sanity#0702ab07e332

mathiazom:
Closed by #580

PR #580, opened 2024-09-04
v3 - basic sitemap generation

mathiazom:
Further explanation is found here: #578 

Automatic sitemap generation based on Sanity documents with a defined `slug` value. Sitemap is available to crawlers at `/sitemap.xml` and the full url is included in a generated `robots.txt`.

We are assuming that all documents containing the field `slug` should be included in the sitemap. If this assumption is wrong, the GROQ query can be adjusted to for example only include specific document types (like blog and page).

Added environment value `NEXT_PUBLIC_URL` for development and production, as this is required to provide fully qualitifed URLs in the sitemap.

image

---

## Checklist


christinaroise:
do/should we cover images in sitemap for variant?
https://developers.google.com/search/docs/crawling-indexing/sitemaps/image-sitemaps

mathiazom:
> do/should we cover images in sitemap for variant?

Next.js sitemap generator was recently added in a canary release (https://github.com/vercel/next.js/releases/tag/v15.0.0-canary.107). So unless it's a top priority we might want to wait for that to be available via the route handler. We might want to test if our images are already available to search engines (e.g. not using JavaScript to load), which would reduce the need to include them in the sitemap.

mathiazom:
TODO: handle pages under subpaths (e.g. blog post at `/blog/hello`)

src/app/sitemap.ts
```diff
@@ -0,0 +1,17 @@
...
+import { client } from "../../studio/lib/client";
+import { Slug } from "../../studio/lib/payloads/global";
+
+interface SitemapDocument {
+  slug: Slug;
+  _updatedAt: string;
+}
+
+export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
+  const documents = await client.fetch<SitemapDocument[]>(`*[defined(slug)]`);
```
christinaroise, line 11: this client fetches without a token — we're certain that's not needed?

christinaroise (approved):

christinaroise:
>Image support in the Next.js sitemap generator was recently added in a canary release (https://github.com/vercel/next.js/releases/tag/v15.0.0-canary.107). So unless it's a top priority we might want to wait for that to be available. We could test if our images are already available to search engines (e.g. not using JavaScript to load), which would reduce the need to include them in the sitemap.

could also be that most of our images are for illustrations only, and that images for say news and costumer cases are more important. But yeah, I agree. Let's hold images until we've got more of the project defined.

mathiazom:
> TODO: handle pages under subpaths (e.g. blog post at `/blog/hello`)

This should be solved when blog post routes are migrated to shared studio, see issue #592

src/app/sitemap.ts
mathiazom: Good catch! This worked for me because my dataset was public. Fixed now to include the token.

PR #582, opened 2024-09-04
v3 - singular form of company location id

mathiazom:
The id used for the company location document was in plural. This is confusing as there is also a field accepting multiple locations using the plural form. The location name of the location document has been given a separate id.

## Checklist

### DIFF
--- a/studio/schemas/documents/companyLocations.ts
+++ b/studio/schemas/documents/companyLocations.ts
@@ -1,16 +1,16 @@
 import { defineField, defineType } from "sanity";
 
-export const companyLocationsID = "companyLocations";
 export const companyLocationID = "companyLocation";
+export const companyLocationNameID = "companyLocationName";
 
 const companyLocations = defineType({
-  name: companyLocationsID,
+  name: companyLocationID,
   type: "document",
   title: "Location",
   description: "Content related to an individual location within the company",
   fields: [
     defineField({
-      name: companyLocationID,
+      name: companyLocationNameID,
       type: "string",
       title: "Location",
     }),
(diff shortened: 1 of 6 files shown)


studio/schemas/documents/companyLocations.ts
```diff
@@ -1,16 +1,16 @@
 import { defineField } from "sanity";
 
-export const companyLocationsID = "companyLocations";
 export const companyLocationID = "companyLocation";
+export const companyLocationNameID = "companyLocationName";
 
 const companyLocations = defineField({
```
anemne, line 6: If we change name to companyLocation, should not classname and function change as well?

mathiazom: Yep, thought I did that, will fix :+1:

anemne (approved):

christinaroise (commented):
the document companyLocations is a list of companyLocations.. hence plural.

mathiazom:
> the document companyLocations is a list of companyLocations.. hence plural.

No, the document itself only represents a single location. However, multiple company locations can be defined via Studio because of `documentTypeList(companyLocationID)`. This is the same as legal documents and dynamic pages. The field `locations` in `studio/schemas/objects/locations.ts` can be plural because that is a field for an array of multiple locations.

christinaroise:
> No, the document itself only represents a single location. However, multiple company locations can be defined via Studio because of documentTypeList(companyLocationID). This is the same as legal documents and dynamic pages. The field locations in studio/schemas/objects/locations.ts can be plural because that is a field for an array of multiple locations.

Good point. Perhaps we should add some cleanup in this pr

christinaroise:
...or in a separate PR. I think that when we notice smaller cleanups like this, we should adjust for the entire repo so to keep the consistency

christinaroise (approved):

PR #617, opened 2024-09-11
v3 - show salary for selected location, degree and examination year (web)

mathiazom:
Further explanation is found here: #616 

Replaces the hardcoded payscale with the location-based salary from Sanity.

No visual changes

## Checklist


src/compensations/Compensations.tsx
```diff
@@ -34,6 +40,16 @@ const Compensations = ({ compensations, locations }: CompensationsProps) => {
...
+    () =>
+      salariesFromLocation(
+        currentYear,
+        selectedLocation,
+        compensations.salaries,
+      ),
+    [currentYear, selectedLocation, compensations.salaries],
+  );
+
   const updateSelectedDegree = (newDegree: Degree) => {
```
anemne, line 53: Is it possible to update updateSelectedDegree() and updateExaminationYear() with one shared method for updating the formValues?

src/compensations/utils/salary.ts
```diff
@@ -0,0 +1,54 @@
...
+  degree: string,
+  salaries: Salaries,
+): number | undefined {
+  const degreeValue = degree === "bachelor" ? 1 : 0;
+  const adjustedYear = examinationYear + degreeValue;
+  return salaries[adjustedYear];
+}
+
+export function calculatePension(salary: number): number {
+  return Math.round(salary * 0.07);
```
anemne, line 19: We have a pension field in sanity as well. Maybe we should update this to use that as well

src/compensations/utils/salary.ts
```diff
@@ -0,0 +1,54 @@
...
+  return Math.min(...examinationYearsFromSalaries(salaries));
+}
+
+export function maxSalariesExaminationYear(salaries: Salaries): number {
+  return Math.max(...examinationYearsFromSalaries(salaries));
+}
+
+export function salariesFromLocation(
+  year: number,
+  locationId: string,
```
anemne, line 36: Better to use location reference instead of this?

src/compensations/utils/salary.ts
mathiazom: Yep, can do that in a separate PR :+1:

src/compensations/Compensations.tsx
mathiazom: `setFormState` is kind of the shared method, while `updateSelectedDegree` and `updateExaminationYear` are used to avoid inlining function bodies in the `<SalaryCalculator />` props

src/compensations/utils/salary.ts
mathiazom: We could, but the `locations` given to `<Compensations />` are of type `CompanyLocation`, not `Reference`. So we would need to convert `CompanyLocation` to `Reference`, i.e. adding `_type`, and then not use `_type` (just `_ref`) 🙃

christinaroise: Perhaps we need some documentation here to avoid confusion for other devs?

src/compensations/utils/salary.ts
```diff
@@ -0,0 +1,54 @@
...
+  salariesByLocation: SalariesByLocation[],
+): Result<Salaries, string> {
+  const yearlySalaries = salariesByLocation.find(
+    (s) => s.location._ref === locationId,
```
christinaroise, line 40: I'm a big fan of spelling it out
```suggestion
    (salary) => salary.location._ref === locationId,
```

studio/components/salariesInput/utils/parseSalaries.tsx
```diff
@@ -4,6 +4,16 @@ export interface Salaries {
   [year: string]: number;
 }
 
+export function isSalariesType(o: unknown): o is Salaries {
```
christinaroise, line 7: same

```suggestion
export function isSalariesType(object: unknown): object is Salaries {
```

studio/lib/payloads/compensations.ts
```diff
@@ -31,5 +38,6 @@ export interface CompensationsPage {
   page: string;
   slug: Slug;
   benefitsByLocation: BenefitsByLocation[];
+  salaries: SalariesByLocation[];
```
christinaroise, line 41: I see we're doing salaries here, but benefitsByLocation above. Let's do either or

mathiazom: That requires a breaking schema change, do we want to do it in this PR?

christinaroise: No not at all. We can do it in a cleanup?

Issue #611, opened 2024-09-11
Fixed sorting in objects array field


mathiazom:
There does not seem to be any way to fix the ordering of objects added to an array field. The only alterantive we know of is to use the `sortable` option, which only toggles user custom sorting. 

A use case would be to sort location bonuses by year. This can be done manually by the user if `sortable=true`, but it would be nice to have this by default with `sortable=false`.

![Image](https://github.com/user-attachments/assets/2d0fa9a7-6e9a-4157-8e82-0eb46b8be401)

Related issue on Sanity repository: https://github.com/sanity-io/sanity/issues/4875

mathiazom:
Closed by #683

PR #683, opened 2024-09-19
v3 - custom ordered array input component

mathiazom:
See #611 

Added custom input compent to allow fixed ordering of array input elements in Studio. Currently only supports direct value comparison, i.e. not references.

Used for ordering salary and bonuses data by year (descending).

image

## Checklist

### DIFF
--- a/studio/components/ValueOrderedArrayOfObjectsInput.tsx
+++ b/studio/components/ValueOrderedArrayOfObjectsInput.tsx
@@ -1,13 +1,14 @@
 import { ArrayOfObjectsInputProps, ObjectItem } from "sanity";
 
-interface OrderedArrayOfObjectsInputProps extends ArrayOfObjectsInputProps {
+interface ValueOrderedArrayOfObjectsInputProps
+  extends ArrayOfObjectsInputProps {
   valueCompareFn: (a: ObjectItem, b: ObjectItem) => number;
 }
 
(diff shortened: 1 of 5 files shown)


studio/components/ValueOrderedArrayOfObjectsInput.tsx
```diff
@@ -0,0 +1,28 @@
+import { ArrayOfObjectsInputProps, ObjectItem } from "sanity";
+
+interface OrderedArrayOfObjectsInputProps extends ArrayOfObjectsInputProps {
```
anemne, line 3: bit heavy to read this 🥲 What about orderedArrayInput or orderedObjectArray

mathiazom: This was to emphasize that it's an extension of the original `ArrayOfObjectsInputProps`. It should actually be even longer:`ValueOrderedArrayOfObjectsInputProps` 🙃 

We need `...Input...` and `...Props` to follow convention and differentiate with the component. So either `OrderedArrayInputProps` or `OrderedObjectArrayInputProps`. The former does not convey that this only covers arrays of objects (compared to primitives), and the latter is barely shorter with an inconsistent naming.

I prefer giving descriptive and fully qualified names by default, and then shorten them if they are used in a lot of places.

anemne (approved):

christinaroise (approved):

Issue #694, opened 2024-09-20
SEO does not work for non-dynamic pages (aka pageBuilder)


mathiazom:

mathiazom:
Closed by #697

PR #697, opened 2024-09-20
v3 - include non-dynamic pages in SEO fetching

mathiazom:
See #694 

Currently, SEO data is not fetched for non-dynamic pages. This is becuase the query has a filter on the `pageBuilder` type. This has been changed to filter based on the existence of the `seo` field instead. This means that pages such as `Compensations` and `CustomerCasePage` also get SEO metadata.

`CustomerCasePage` was missing the seo fields, so this has also been added.

## Checklist

### DIFF
--- a/studio/lib/queries/page.ts
+++ b/studio/lib/queries/page.ts
@@ -77,7 +77,7 @@ export const SLUG_QUERY = groq`
 `;
 
 export const SEO_SLUG_QUERY = groq`
-  *[_type == "pageBuilder" && slug.current == $slug][0]{
+  *[defined(seo) && slug.current == $slug][0]{
         "title": seo.seoTitle,
         "description": seo.seoDescription,
         "imageUrl": seo.seoImage.asset->url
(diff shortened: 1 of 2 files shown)


studio/lib/queries/page.ts
```diff
@@ -77,7 +77,7 @@ export const SLUG_QUERY = groq`
 `;
 
 export const SEO_SLUG_QUERY = groq`
-  *[_type == "pageBuilder" && slug.current == $slug][0]{
+  *[defined(seo) && slug.current == $slug][0]{
```
christinaroise, line 80: how would this transfer for documents where we don't have or want seo? I'm assuming the default SEO would then be the fallback? Say for Legal documents

christinaroise (approved):

anemne (approved):

studio/lib/queries/page.ts
christinaroise: perff

Issue #699, opened 2024-09-20
Support image block in shared richtext


mathiazom:
Images in rich text from shared are not rendered. This is becuase the image assets are fetched with the studio client, not the shared client.

mathiazom:
A simple solution would be to attempt to fetch with both studio and shared client, and pick whichever it not 404. This assumes that the asset ids are globally unique. 

A more involved, but possibly safer, approach is to pass either a client identifier or the client config itself through the RichText hierarchy. This would require some prop drilling though 👷

christinaroise:
How many layers of prop drilling are we looking at? 🫠

mathiazom:
> How many layers of prop drilling are we looking at? 🫠

- `RichText`
  - `GroupedBlogText`
    - `myPortableTextComponents` (convert to function)
      - `SanityImage`
        - `useConvertSanityImageToNextImage`
          - `SanityImage`

christinaroise:
Hm, could perhpas the RichText component in and of itself be optimized? 🤔 there are also cases where we have myPortableTextComponents because RichText doesnt support normal as any other styles than body, so for cases where we want portable text but with normal as say bodySuperLarge or bodyLarge.. 🤔

mathiazom:
Closed by #702

PR #702, opened 2024-09-23
v3 - support image assets from shared Sanity project

mathiazom:
See #699 

Added `useNextSanityGlobalImage` to search both studio and shared when rendering images. Also refactored the use of image hooks and rendering.

Before | **After**
--- | ---
image | image

Storybook should not be affected, but a fix was added to add the required width and height props to Next.js Image:

image

## Checklist


anemne (approved):

christinaroise:
could you add screenshots of how this affects stories?

mathiazom:
> could you add screenshots of how this affects stories?

Should not affect stories, but they were actually broken already... Added a fix now.

src/components/image/SanityImage.tsx
```diff
@@ -0,0 +1,78 @@
...
+  );
+};
+
+export function SanityImage({ image }: { image: IImage }) {
+  if (image?.src) {
+    return (
+      <Image
+        alt={image?.alt || ""}
+        src={image.src.src}
+        style={{ objectFit: "cover", height: "100%" }}
```
christinaroise, line 71: Discovered that this fix might be needed for a few cases

```suggestion
        style={{ objectFit: "cover", height: "100%", width: "100%"  }}
```

christinaroise: could then be that you don't need the width and height props below, but stories might want them anyway.. 🤷

christinaroise (approved):

src/components/image/SanityImage.tsx
mathiazom: `width` and `height` are required props for [Next.js Image](https://nextjs.org/docs/app/api-reference/components/image), but we can add `width: "100%"` if it's needed :+1:

PR #703, opened 2024-09-23
v3 - update language schema and add doc translation

christinaroise:
## Short Description

This feature introduces a custom language component to manage default and supported languages within the main studio. By default, the LanguageSettings schema initializes with English.

The [documentInternationalization](https://github.com/sanity-io/document-internationalization) plugin currently lacks support for asynchronous language lists in initial value templates. As a result, all supported languages are displayed in the studio.

Users can set a default language and add supported languages, establishing a reliable reference for the frontend. However, this feature does **not** filter content based on the selected supported languages. Meaning, this allows users to create content in languages not supported by the site. For example, while Variant Norge may select Norwegian and English, Swedish will still appear as a translation option.

To ensure data integrity, documents without a language value are filtered out, preventing the creation of documents that lack a specified language. This design also facilitates future adjustments for integrating asynchronous language lists with minimal effort.

---

Language Settings
Screenshot 2024-09-19 at 21 39 08

Document translation has been added to Legal Documents, they are currently not sorted.
Screenshot 2024-09-23 at 15 21 46

---

## Checklist

- [ ] **Documentation**: Relevant documentation has been added or updated (if applicable). 
I'm creating an issue to update the readme with the translations setup once we have field translation up and running too.


internationalization/supportedLanguages.ts
```diff
@@ -0,0 +1,31 @@
...
+  _key: string;
+  title: string;
+  id: string;
+  icon: string;
+  default?: boolean;
+}
+
+export const supportedLanguages: Language[] = [
+  {
+    _key: `${Date.now()}-${Math.random().toString(36).slice(2, 11)}`,
```
mathiazom, line 11: Any specific reason to use this id? Could we use UUID instead?

internationalization/supportedLanguages.ts
```diff
@@ -0,0 +1,31 @@
...
+    title: "English",
+    id: "en",
+    icon: "🇬🇧",
+    default: true,
```
mathiazom, line 15: Can we have a separate value for default language? This avoids multiple defaults, and would be easier to use directly (avoid `.find()`)

src/app/layout.tsx
```diff
@@ -35,15 +35,14 @@ export default async function RootLayout({
   let siteLang;
 
   try {
-    const { data } = await loadStudioQuery<DefaultLanguageObject>(
+    const { data } = await loadStudioQuery<LanguageObject[]>(
       DEFAULT_LANGUAGE_QUERY,
     );
-    siteLang = data.defaultLanguage;
+    siteLang = data[0].id;
```
mathiazom, line 41: Can we add `[0]` to the GROQ query instead of here?

studio/deskStructure.ts
```diff
@@ -56,7 +55,7 @@ const adminSection = (S: StructureBuilder) =>
             .title("Legal Documents")
             .icon(CogIcon)
             .child(
-              S.documentTypeList(legalDocumentID).title("Legal Documents"),
+              S.documentTypeList(legalDocumentID).title("Company Locations"),
```
mathiazom, line 58: This looks wrong 🙃

studio/schemas/documents/languageSettings.ts
```diff
@@ -0,0 +1,37 @@
+"use client";
```
mathiazom, line 1:
```suggestion
```

internationalization/supportedLanguages.ts
christinaroise: nope just me forgetting to add the import 🙃

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,151 @@
...
+  Language,
+  supportedLanguages,
+} from "internationalization/supportedLanguages";
+
+interface LanguageSelectorProps {
+  value?: Language[];
+  onChange: (event: PatchEvent) => void;
+}
+
+const colorMap = {
```
anemne, line 15: Should this be moved to a shared utils/styling file?

anemne (approved):
Looks good!

studio/components/LanguageSelector.tsx
christinaroise: Hm good question. This is more specific to this selector, but it's definitely worth looking into in the issue created for shared color values.

studio/schemas/documents/admin/legalDocuments.ts
```diff
@@ -1,18 +1,35 @@
...
   title: "Legal Document",
-  fields: [title, titleSlug, richText],
+  fields: [languageSchemaField, title, titleSlug, richText],
   preview: {
     select: {
       title: "basicTitle",
+      language: "language",
+    },
+    prepare({ title, language }) {
+      const languageEntry = supportedLanguages.find(
```
mathiazom, line 21: Could be nice to extract this to a common function (e.g. `getLanguageById`) in `supportedLanguages.ts`

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  return (
+    <Flex direction="column" gap={3}>
+      {supportedLanguages.map((lang) => (
+        <Card
+          key={lang._key}
+          padding={4}
+          radius={2}
+          shadow={1}
+          style={{
```
mathiazom, line 82: Should we not avoid inline styling?

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const prefersDark = theme.sanity.v2?.color._dark ?? false;
+
+  // Get the currently set default language
+  const currentDefaultLanguage = value.find((lang) => lang.default)?.id || null;
+
+  const handleLanguageSelection = (lang: Language) => {
+    const isSelected = value.some((item) => item.id === lang.id);
+    // Prevent deselecting the last remaining language
+    if (isSelected && value.length === 1) {
+      console.log("Cannot deselect the last remaining language.");
```
mathiazom, line 37: Should we use `console.log` in production?
```suggestion
```

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+import {
+  Language,
+  supportedLanguages,
+} from "internationalization/supportedLanguages";
+
+interface LanguageSelectorProps {
+  value?: Language[];
+  onChange: (event: PatchEvent) => void;
+}
```
mathiazom, line 13:
```suggestion
type LanguageSelectorProps = ArrayOfObjectsInputProps<Language>;
```

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,151 @@
...
+      currentDefaultLanguage,
+      lang,
+    );
+
+    const finalValue = updatedValue.map((item) => ({
+      ...item,
+      default: item.id === newDefaultLanguage,
+    }));
+
+    onChange(PatchEvent.from(set(finalValue)));
```
mathiazom, line 56:
```suggestion
    if (isSelected && lang.id === currentDefaultLanguage) {
      // default was deselected, set first language as default
      updatedValue[0].default = true;
    }

    onChange(PatchEvent.from(set(updatedValue)));
```

studio/components/LanguageSelector.tsx
christinaroise: nope, this was just left over from debugging

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,151 @@
...
+            backgroundColor: getBackgroundColor(lang),
+          }}
+          onClick={() => handleLanguageSelection(lang)}
+        >
+          <Flex align="center" justify="space-between">
+            <Flex align="center">
+              <Checkbox
+                id={lang.id}
+                checked={value.some((item) => item.id === lang.id)}
+                readOnly
```
mathiazom, line 93: Should be able to use the actual checkbox as well. Also made it disabled if only language left.
```suggestion
                onChange={() => handleLanguageSelection(lang)}
                disabled={
                  value?.length <= 1 && currentDefaultLanguage === lang.id
                }
```

studio/components/LanguageSelector.tsx
christinaroise: I've been strict for plain html elements, but not for sanity ui. 🤔  we could do a cleanup when we look at the css variables for colors in studio as well

studio/components/LanguageSelector.tsx
christinaroise: The original doesn't set the first language as the default but the next selected, so if 2 and 3 is selected where 2 is default, and I unselect 2, number 3 is set as default.

studio/components/LanguageSelector.tsx
christinaroise: Small Sanity issue here. If we do add onChange we'll also need an `e.stopPropagation(); // Prevent triggering card's click handler` due to the onClick on parent level. However, this disables on click on the label itself. 

If we remove the onClick on the card the checkbox and the label is triggered..

studio/components/LanguageSelector.tsx
mathiazom: Is that important? Does the user expect this behaviour, or would they equally accept the new default as the first language.

studio/components/LanguageSelector.tsx
christinaroise: Opting for the original solution with readOnly, as the little checkbox covers a much smaller unclickable area than the label itself.

studio/components/LanguageSelector.tsx
mathiazom: I don't the original has that behaviour either. Select 1,2 and 3, mark 2 as default and deselect 2. You would then expect 3 to be set as default, but it's actually 1.

mathiazom: The fact that the languages are not necessarily in the same order in the selection list as in `supportedLanguages` makes it a little hard to reason about this behaviour. I think it would be clearer if we look at only the IDs when selecting and deselecting, and then rebuild the selection array by filtering `supportedLanguages` with the selected IDs to preserve the same order as `supportedLanguages`.

christinaroise: Yes, that's correct, it takes the first selected object in the array. We could ofc take the next in line, but I don't know if that's a good experience.  I gave it a try and found it a little confusing at the time

And we do want that UX for cases where you've chosen 2 and 3, but deselect 2, you don't want 1 to pop up as selected with default.

mathiazom: The fact that the languages are not necessarily in the same order in the selection list as in supportedLanguages makes it a little hard to reason about this behavior. I think it would be clearer if we look at only the IDs when selecting and deselecting, and then rebuild the selection array by filtering supportedLanguages with the selected IDs to preserve the same order as supportedLanguages.

What about something like this?

```ts
const handleLanguageSelection = (lang: Language) => {
  const selectedIds = value?.map((language) => language.id);
  const isSelected = selectedIds?.includes(lang.id);
  // Prevent deselecting the last remaining language
  if (isSelected && selectedIds.length === 1) {
    return; // Exit early if there's only one language selected
  }
  const updatedSelectedIds = isSelected
    ? selectedIds.filter((id) => id !== lang.id) // Deselect language
    : [...selectedIds, lang.id]; // Select language
  const needNewDefault =
    (isSelected && lang.id === currentDefaultLanguage) ||
    currentDefaultLanguage === null;
  const updatedValue = supportedLanguages.reduce<Language[]>(
    (acc, language) => {
      if (!updatedSelectedIds.includes(language.id)) {
        return acc;
      }
      return [
        ...acc,
        {
          ...language,
          default: needNewDefault
            ? acc.length === 0
            : currentDefaultLanguage === language.id,
        },
      ];
    },
    [],
  );
  onChange(PatchEvent.from(set(updatedValue)));
};
```

christinaroise: Why do you want value (selected languages) to be the same order as supportedLanguages?

mathiazom: So that the new default does not depend on the order in which you selected the languages.

christinaroise: Ah, gotcha!

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const handleDefaultSetting = (lang: Language) => {
+    const updatedValue = value.map((item) => ({
+      ...item,
+      default: item.id === lang.id, // Only one language can be the default
+    }));
+    onChange(PatchEvent.from(set(updatedValue)));
+  };
+
+  // Memoize value for performance and reusability
+  const languageCards = useMemo(
```
christinaroise, line 62: somewhat uncertain if useMemo is really needed. bob bob 🤷‍♀️

mathiazom: There are really no expensive operations here, so I think we should drop it for now.

studio/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,127 @@
...
+    if (isSelected && lang.id === currentDefaultLanguage) {
+      const nextDefaultLanguage = getNextDefaultLanguage(updatedValue, lang);
+      updatedValue = updatedValue.map((item) => ({
+        ...item,
+        default: item.id === (nextDefaultLanguage?.id || null),
+      }));
+    }
+
+    onChange(PatchEvent.from(set(updatedValue)));
+  };
```
mathiazom, line 61: Just a suggestion for a cleaner iterator for finding a new default, and an abort if a new default could not be found (edge-case that should not really happen).
```suggestion
const getNextDefaultLanguageId = (
    updatedValue: Language[],
    oldDefaultLanguage: Language,
  ) => {
    const oldIndex = supportedLanguages.findIndex(
      (language) => language.id === oldDefaultLanguage.id,
    );
    for (let i = 0; i < supportedLanguages.length; i++) {
      const languageId =
        supportedLanguages[(i + oldIndex) % supportedLanguages.length].id;
      if (updatedValue.some((item) => item.id === languageId)) {
        return languageId;
      }
    }
    return null;
  };

  const handleLanguageSelection = (lang: Language, isSelected: boolean) => {
    if (isSelected && value.length === 1) return;

    let updatedValue = isSelected
      ? value.filter((item) => item.id !== lang.id)
      : [...value, { ...lang, default: false }];

    if (isSelected && lang.id === currentDefaultLanguage) {
      const nextDefaultLanguageId = getNextDefaultLanguageId(
        updatedValue,
        lang,
      );
      if (nextDefaultLanguageId === null) {
        // abort selection change
        return;
      }
      updatedValue = updatedValue.map((item) => ({
        ...item,
        default: item.id === nextDefaultLanguageId,
      }));
    }

    onChange(PatchEvent.from(set(updatedValue)));
  };
```

mathiazom (approved):

PR #716, opened 2024-09-26
query field translations + richtext

christinaroise:
Updated GROQ queries to take language into account. This way we only fetch the translation for the current language, and web components don't have to handle unpacking internationalized strings. The downside is added complexity in the GROQ queries. This is somewhat alleviated by defining reusable query "fragments".

Examples:
```typescript
export const TRANSLATED_LINK_FRAGMENT = groq`
    ${LANGUAGE_FIELD_FRAGMENT},
    ${TRANSLATED_LINK_TITLE_FRAGMENT},
    ${TRANSLATED_INTERNAL_LINK_FRAGMENT}
`;

const SECTIONS_FRAGMENT = groq`
  sections[]{
    ...,
    _type == "hero" => {
      ...,
      callToActions[] {
        ...,
        ${TRANSLATED_LINK_FRAGMENT}
      }
    },
    _type == "article" => {
      ...,
      link {
        ...,
        ${TRANSLATED_LINK_FRAGMENT}
      }
    },
    _type == "callout" => {
      ...,
      link {
        ...,
        ${TRANSLATED_LINK_FRAGMENT}
      }
    },
    _type == "ctaSection" => {
      ...,
      callToActions[] {
        ...,
        ${TRANSLATED_LINK_FRAGMENT}
      }
    }
  }
`;
```

```typescript
export function translatedFieldFragment(fieldName: string) {
  return groq`${fieldName}[_key == $language][0].value`;
}

// ...

export const COMPENSATIONS_PAGE_QUERY = groq`
  *[_type == "compensations" && ${translatedFieldFragment("slug")} == $slug][0] {
    ...,
    "language": $language,
    "slug": ${translatedFieldFragment("slug")},
    "basicTitle": ${translatedFieldFragment("basicTitle")},
    "benefitsByLocation": benefitsByLocation[] {
      ...,
      "benefits": benefits[] {
        ...,
        "basicTitle": ${translatedFieldFragment("basicTitle")},
        "richText": ${translatedFieldFragment("richText")}
      }
    },
    "seo": ${translatedFieldFragment("seo")}
  }
`;
```

As an example on how to add translations to richText, this PR adds richtext translations in benefit

Translated rich text in benefits:
Screenshot 2024-09-26 at 09 55 12

## Checklist


anemne:
You should add to PR description that you have added localized value for title as well.

mathiazom:
"Hijacked" this PR to also include handling of field translations in GROQ queries.

src/app/(main)/legal/[id]/page.tsx
```diff
@@ -20,7 +20,7 @@ async function Page({ params }: Props) {
   const { perspective, isDraftMode } = getDraftModeInfo();
 
   const initialDocument = await loadStudioQuery<LegalDocument>(
-    LEGAL_DOCUMENTS_BY_SLUG_AND_LANG_QUERY,
+    LEGAL_DOCUMENT_BY_SLUG_AND_LANG_QUERY,
     { slug: id, language: "en" }, //TODO: replace this with selected language for the page
```
Comment (anemne) [inline src/app/(main)/legal/[id]/page.tsx:24]: We could probably update this hard coded variable as well.

Comment (mathiazom) [reply src/app/(main)/legal/[id]/page.tsx:24]: This file will be removed by #734 , so didn't bother to update it here 🙃

Comment (anemne) [reply src/app/(main)/legal/[id]/page.tsx:24]: You're right!

studio/schemas/objects/link.ts
```diff
@@ -190,10 +192,12 @@ export const link = defineField({
       title: "linkTitle",
       type: "linkType",
     },
-    prepare(selection) {
-      const { title, type } = selection;
+    prepare({ title, type }) {
+      const translatedTitle = isInternationalizedString(title)
```
christinaroise, line 196: Because we know that link is of type  `internationalizedArrayString`, we don't need this check.

studio/schemas/objects/link.ts
```diff
@@ -190,10 +192,12 @@ export const link = defineField({
...
+        ? firstTranslation(title)
+        : null;
       return {
-        title: title[0].value,
+        title: translatedTitle ?? "Missing title",
```
christinaroise, line 200:
``` suggestion
        title: title[0].value || null,
```

studio/lib/interfaces/global.ts
```diff
@@ -8,9 +8,26 @@ export interface DocumentWithSlug {
   _updatedAt: string;
 }
 
-export type InternationalizedString = InternationalizedStringValue[];
+export function isInternationalizedString(
```
christinaroise, line 11: this is a function but placed inside the interfaces folder?

studio/lib/queries/pages.ts
```diff
@@ -1,87 +1,105 @@
 import { groq } from "next-sanity";
 
-// TODO: set all sectons.ref's to slug.current
+export const LANGUAGE_FIELD_FRAGMENT = groq`
+  "language": $language
+`;
+
+export function translatedFieldFragment(fieldName: string) {
```
christinaroise, line 7: same here. move to a util file instead of placing them under queries??

I see why you've placed it here, but because it's a function it doesn't really fit here.

studio/lib/queries/pages.ts
```diff
@@ -1,87 +1,105 @@
...
+  return groq`${fieldName}[_key == $language][0].value`;
```
christinaroise, line 8: Why have you chosen this approach instead of: 

```
  // "basicTitle": basicTitle[$language],
  // "slug": slug[$language].current,
 ```

Comment (christinaroise) [reply studio/lib/queries/pages.ts:8]: I think, if possible, the suggestion above would give us less bloat and a cleaner approach.

**Review by christinaroise — COMMENTED** [review]
make sure you also update the PR description

File: studio/lib/interfaces/global.ts (hunk 4 above)
Comment (mathiazom) [reply studio/lib/interfaces/global.ts:11]: This is a type guard for the type defined in this file. So I valued closeness to the relevant type over having only types/interfaces in the file.

File: studio/schemas/objects/link.ts (hunk 2 above)
Comment (mathiazom) [reply studio/schemas/objects/link.ts:196]: We don't actually know that though. More importantly, TypeScript doesn't know what type it is (since `prepare` gives us variables of type `any`). With the type guard, we can safely ignore the value if it doesn't meet our expectations, and otherwise use the variable with proper typing (which makes the use of `firstTranslation` much nicer). Without the type guard, we are using array operations on something that might not be an array and accessing properties that might not exist.

This is of course a question of how "safe" we want to be, but since we have favored the safer side [before](https://github.com/varianter/variant.no/pull/535#discussion_r1744997509) I think this check fits nicely.

File: studio/schemas/objects/link.ts (hunk 3 above)
Comment (mathiazom) [reply studio/schemas/objects/link.ts:200]: See comment above, but returning `null` is probably better than `"Missing title"` :+1:

File: studio/lib/queries/pages.ts (hunk 5 above)
Comment (mathiazom) [reply studio/lib/queries/pages.ts:7]: Will move this to a util file, but mostly because it will be reused in other query files. In general, I don't think we should split up related concepts just because they are functions or variables.

File: studio/lib/queries/pages.ts (hunk 6 above)
Comment (mathiazom) [reply studio/lib/queries/pages.ts:8]: Because that's the query required by the shape of the data: https://github.com/sanity-io/sanity-plugin-internationalized-array/blob/main/README.md#shape-of-stored-data

File: studio/schemas/objects/link.ts (hunk 2 above)
Comment (christinaroise) [reply studio/schemas/objects/link.ts:196]: Why don't we know it's of type internationalizedArrayString when that's why we're setting on the schema? 

I agree however that there should be a better approach given cases where we update the schema type but forget to adjust in preview. Could we leave it as it was and come up with a better approach for all preview cases? Because this wouldn't just concern string array, but cases where we say adjust the schema name that's used as reference in a preview

File: studio/lib/queries/pages.ts (hunk 6 above)
Comment (christinaroise) [reply studio/lib/queries/pages.ts]: Ah my bad, I was looking ad doc translation 💡 

I am still concerned however with the amount of bloat and complexity this approach causes. I'd vote for opting out of the added fragments and if absolutely needed we could go back and revisit.

File: studio/schemas/objects/link.ts (hunk 2 above)
Comment (mathiazom) [reply studio/schemas/objects/link.ts:196]: > Why don't we know it's of type internationalizedArrayString when that's why we're setting on the schema?

Because there is no typing of the preview variables. So the schema could be changed without raising any errors. With the type guard we can try to give the variable a proper type, and safely fail if the schema has changed. Using a variable with an actual type is much easier than using an `any` variable, which could be anything. We could also use the `as` keyword to achieve this, but then we would just bypass the type check completely and possibly end up with undefined behavior. 

> I agree however that there should be a better approach given cases where we update the schema type but forget to adjust in preview. Could we leave it as it was and come up with a better approach for all preview cases? Because this wouldn't just concern string array, but cases where we say adjust the schema name that's used as reference in a preview

This is an attempt at checking that. I agree that this check would be nice in other uses of `prepare`, but I don't see why that should stop us from doing it only here as a start.

**Review by anemne — APPROVED** [review]

File: studio/lib/queries/pages.ts (hunk 6 above)
Comment (mathiazom) [reply studio/lib/queries/pages.ts]: The whole point of using fragments is to reduce bloat and address the complexity introduced in queries when using translated strings. If we don't use fragments, we have to repeat the same GROQ expressions in all the places we have translations. This increases the chance that we make mistakes when writing queries.

Comment (christinaroise) [reply studio/lib/queries/pages.ts]: I agree, I just think that for cases where there's only one line, such as `basicTitle[_key == $language][0].value` there's no need for fragments

File: studio/schemas/objects/link.ts (hunk 2 above)
Comment (christinaroise) [reply studio/schemas/objects/link.ts:196]: My concern is that this approach is too specific to this scenario and that it wouldn't transfer nicely in other scenarios, giving us multiple type guards. The main problem is that our checks aren't triggered if preview get's an error. How about we start there first before we add type guards?

File: studio/lib/queries/i18n.ts [hunk 7]
```diff
@@ -0,0 +1,28 @@
+import { groq } from "next-sanity";
```
christinaroise, line 1: our root folder for translation is named `internationalization` let's rename this file to the same to more easily understand the connection

studio/lib/queries/pages.ts
christinaroise: Let me clarify further. The Link fragments are absolutely ok! 
But I think the `translatedFieldFragment()` is redundant

src/blog/components/legal/LegalPreview.tsx
```diff
@@ -2,7 +2,7 @@
 import { QueryResponseInitial, useQuery } from "@sanity/react-loader";
```
christinaroise, line 2: Remind me why we added `app/blog/components` ?

studio/lib/queries/i18n.ts
mathiazom: Would it be possible to rename the root folder to `i18n` instead? Internationalization is a famously long word, and "i18n" is widely adopted 🙃

src/blog/components/legal/LegalPreview.tsx
mathiazom: I don't see that folder, are you talking about `src/blog/components`? If not, it might be a folder on your local machine that is not tracked by git (since git only tracks files).

studio/lib/queries/pages.ts
mathiazom: Ok, good 🤗 

Based on your initial comment, I think we can agree that `[_key == $language][0].value` is not very readable or clean. It's also used many places, so reusing the expression makes sense. `translatedFieldFragment()´ is therefore both an attempt at demystifying the expression and to enable reuse.

studio/schemas/objects/link.ts
mathiazom: I don't see why we shouldn't use a type guard just because it doesn't solve all our problems related to preview. What is your concern with multiple type guards? A type guard is only for one type, so naturally you would need multiple type guards to handle different types.

Are you maybe thinking that we should throw an error instead of failing silently? We could still do that if the type guard returns false.

src/blog/components/legal/LegalPreview.tsx
christinaroise: Sorry my bad, it looked like it was under app.

Still wondering why we've got `blog`?

studio/lib/queries/i18n.ts
christinaroise: hm, I suppose. It's just that Internationalization is the term when adding translations to websites or apps. 🤔  But if the two of you think i18n is sufficient enough without having to google it. Go ahead 👍

studio/lib/queries/pages.ts
christinaroise: No, that's not what I meant. 

I meant that `[_key == $language][0].value` is sufficient enough and to omit the use of `translatedFieldFragment()`

studio/schemas/objects/link.ts
christinaroise: My concern is that the handling of previews are done manually right now. There are no checks or tests that fail if preview throws an error such as `Invalid Preview Config`. So if a dev ads changes to the schema, and specifically changes schema values that are used in preview, we have to rely on our memory to update after changes.

So a type guard is nice and all as it gives us visually a fallback, but we're still not alerted to the changes done which preview has to account for

src/blog/components/legal/LegalPreview.tsx
mathiazom: The folder has been there since the beginning, so you tell me 🙃

studio/lib/queries/pages.ts
mathiazom: I don't agree that it's redundant, seeing as we already use it nine different places with just these initial translations. However, I'm fine with removing it if it means we can get this merged.

src/blog/components/legal/LegalPreview.tsx
christinaroise: Ahh right. So then this is placed in a wrong folder. 

blog was meant for UI specific blog and blogpost content.

studio/schemas/objects/link.ts
mathiazom: Added a TypeError that throws if the type guard fails. It will then display an error in the preview component.

image

studio/lib/queries/pages.ts
mathiazom: I will leave this as is for now. We can always inline everything later if it becomes an issue.

src/blog/components/legal/LegalPreview.tsx
mathiazom: We'll fix this later.

PR #717, opened 2024-09-26
V3 fix render issue in legal documents

anemne:
## Short Description

I’ve added a check to determine if the rich text is undefined. If it is, the user will be informed that it is empty, and added a redirectbutton to the given document. Additionally, I’ve made the rich text field in the legal document required.

Screenshot 2024-09-26 at 11 07 14

Screenshot 2024-09-26 at 11 07 24

### DIFF
--- a/src/blog/components/legal/Legal.tsx
+++ b/src/blog/components/legal/Legal.tsx
@@ -21,7 +21,6 @@ const extractHeadings = (blocks: PortableTextBlock[]) => {
 };
 
 const Legal = ({ document }: { document: LegalDocument }) => {
-  console.log(document.richText);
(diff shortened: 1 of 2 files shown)


src/blog/components/legal/Legal.tsx
```diff
@@ -19,7 +21,18 @@ const extractHeadings = (blocks: PortableTextBlock[]) => {
 };
 
 const Legal = ({ document }: { document: LegalDocument }) => {
-  const headings = extractHeadings(document.richText);
+  console.log(document.richText);
```
mathiazom, line 24:
```suggestion
```

mathiazom (approved):

christinaroise (approved):

PR #758, opened 2024-10-10
v3 - purge unused blog components

mathiazom:
We don't have any plans for a custom hosted blog, and probably won't for a while.

This is a start at removing boilerplate that we don't actually need. This PR mostly removes unused React components, and there are still some unused schemas and other util functions left. This is left for another PR. 

Deleted code is not lost forever, but preserved in the git log. This way we can still use it for reference later, without the burden of maintaining it.


src/post/hero/PostHero.stories.tsx
anemne, line 1: Are you sure we don't need this?

mathiazom: Not 100%, but we can always revisit it if it becomes relevant. The point is just to not lock ourselves into the legacy setup.

anemne (approved):

PR #762, opened 2024-10-10
v3 - disable slug change check

mathiazom:
Time to kill some of my own darlings...

We are currently setting the slug field to read-only after publication and making sure that slugs of published documents are not changed.

I think it makes more sense, and avoids a lot of pain, to trust the user that a redirect will be created if the old slug should still work. If that later seems to not be the case, we can reintroduce these constraints.


anemne (approved):
This seems like a good idea for now

PR #765, opened 2024-10-11
v3 - SEO query cleanup

mathiazom:
Cleanup of some parts of the SEO fetching and type definitions

### DIFF
--- a/src/app/(main)/[lang]/page.tsx
+++ b/src/app/(main)/[lang]/page.tsx
@@ -1,28 +1,19 @@
 import { Metadata } from "next";
 
-import CustomErrorMessage from "src/components/customErrorMessage/CustomErrorMessage";
 import InformationSection from "src/components/informationSection/InformationSection";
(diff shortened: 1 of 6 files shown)


src/app/(main)/[lang]/page.tsx
```diff
@@ -82,34 +49,21 @@ const Home = async ({ params }: Props) => {
...
 
-  if (!initialLandingPage.data) {
-    return (
-      <InformationSection
-        title="Landing Page is Missing Content"
-        body={`Your landing page is set, but it looks like there’s no content yet.\n Visit the Studio to start adding content and make your landing page come to life!`}
-        link={pagesLink}
-      />
-    );
-  }
```
Comment (anemne) [inline src/app/(main)/[lang]/page.tsx:99]: What if landingpage is not set, how will this look like for the end user?

Comment (mathiazom) [reply src/app/(main)/[lang]/page.tsx:99]: It will display the "Welcome! Velkommen! Välkommen!" information section that is returned further up here.

anemne (approved):

PR #766, opened 2024-10-11
v3 - initial employees page

mathiazom:
Initial setup for having a dynamic page section for showcasing all our employees.

In case it is not obvious from the last screenshot, only the skeleton setup has been implemented (i.e. section schema, TS types and page fetching). Actually fetching and displaying employees data is left for #767.

image

image

### DIFF
--- a/src/utils/renderSection.tsx
+++ b/src/utils/renderSection.tsx
@@ -166,7 +166,7 @@ const SectionRenderer = ({
   switch (section._type) {
     case "hero":
       return renderHeroSection(
-        section as HeroSection,
+        section,
         sectionIndex,
         isDraftMode,
         initialData,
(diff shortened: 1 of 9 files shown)


src/utils/renderSection.tsx
```diff
@@ -166,60 +166,67 @@ const SectionRenderer = ({
...
         initialData,
       );
     case "grid":
-      return renderGridSection(
-        section as GridSection,
-        sectionIndex,
-        isDraftMode,
-        initialData,
+      return renderGridSection(section, sectionIndex, isDraftMode, initialData);
+    case "employees":
```
anemne, line 219:
```suggestion
    case "employees":
    //Todo: Remove this as the employee section is implemented.
```

src/utils/renderSection.tsx
```diff
@@ -166,60 +166,67 @@ const SectionRenderer = ({
...
+      return (
```
anemne, line 220:
```suggestion
      return (
      //Todo: Replace this as the employee section is implemented.
```

anemne (approved):
👌

PR #767, opened 2024-10-11
v3 - fetch employees data

mathiazom:
Very basic Proof of Concept for fetching external data for the employees page section. 

The section can be added to any dynamic page (aka pageBuilder)

image

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -6,6 +6,10 @@ const nextConfig = {
         protocol: "https",
         hostname: "cdn.sanity.io",
       },
+      {
+        protocol: "https",
+        hostname: "chewiesald2ijhpvmb34c.blob.core.windows.net",
+      },
     ],
   },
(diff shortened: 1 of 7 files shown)


trulshj:
An important concern here is that individual employees should have control over whether or not their phone number should be shown as contact info. This is probably something that should be part of Chewbacca and controlled through dash.variant.no so not something that is directly implementable here, but important to design for this possibility.

mathiazom:
> An important concern here is that individual employees should have control over whether or not their phone number should be shown as contact info. This is probably something that should be part of Chewbacca and controlled through dash.variant.no so not something that is directly implementable here, but important to design for this possibility.

How about filtering this out when constructing the Chewbacca response? We would then only need to handle missing values, i.e. having them as optional in employee type.

anemne (approved):

Issue #592, opened 2024-09-09
Support nested slugs in sitemap generation


mathiazom:
Currently, the sitemap generator only looks at the direct slug when generating absolute urls. This will not work for pages using nested slugs, like the customer cases (e.g. `/kunder/fram` where `fram` is the direct slug, while `kunder` is the parent slug for all customer cases).

mathiazom:
Waiting for functional blog feature via Shared Studio

mathiazom:
Should add detail page for customer cases

PR #776, opened 2024-10-14
v3 - support nested slugs in sitemap

mathiazom:
See #592 

Currently, the sitemap generation only looks at the direct slug for all documents. This does not work for e.g. customer cases, where we have a parent document with multiple child documents (customer cases) requiring two levels of slugs. 

The sitemap has now been updated to produce the full paths, including language code.

image

### DIFF
--- a/src/app/sitemap.ts
+++ b/src/app/sitemap.ts
@@ -1,18 +1,129 @@
 import type { MetadataRoute } from "next";
 
 import { client } from "studio/lib/client";
-import { DocumentWithSlug } from "studio/lib/interfaces/global";
-import { PageBuilder } from "studio/lib/interfaces/pages";
-import { DOCUMENTS_WITH_SLUG_QUERY } from "studio/lib/queries/siteMap";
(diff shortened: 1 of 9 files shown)


src/app/sitemap.ts
```diff
@@ -1,41 +1,149 @@
...
+import { PAGES_SITEMAP_QUERY } from "studio/lib/queries/pages";
+import {
+  LANDING_PAGE_SITEMAP_QUERY,
+  LANGUAGES_QUERY,
+} from "studio/lib/queries/siteSettings";
+import {
+  COMPENSATIONS_PAGE_SITEMAP_QUERY,
+  CUSTOMER_CASES_PAGE_SITEMAP_QUERY,
+} from "studio/lib/queries/specialPages";
 import { token } from "studio/lib/token";
```
anemne, line 21: Is it better to add "token as studiotoken" here?

anemne (approved):

src/app/sitemap.ts
mathiazom: Sure :+1:

PR #778, opened 2024-10-15
V3-translation in customer case page in /studio

anemne:
This PR adds field translation for the customer case page in /studio. 

Related issue: [#771](https://github.com/varianter/variant.no/issues/771)

### DIFF
--- a/studio/lib/queries/specialPages.ts
+++ b/studio/lib/queries/specialPages.ts
@@ -34,7 +34,18 @@ export const COMPENSATIONS_PAGE_SITEMAP_QUERY = groq`
 
 //Customer Cases
 export const CUSTOMER_CASES_PAGE_QUERY = groq`
-  *[_type == "customerCasesPage" && slug.current == $slug][0]`;
+  *[_type == "customerCasesPage" && ${translatedFieldFragment("slug")} == $slug][0]{
+    ...,
+    "language": $language,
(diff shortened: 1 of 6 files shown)


studio/lib/queries/specialPages.ts
```diff
@@ -34,7 +34,18 @@ export const COMPENSATIONS_PAGE_SITEMAP_QUERY = groq`
 
 //Customer Cases
 export const CUSTOMER_CASES_PAGE_QUERY = groq`
-  *[_type == "customerCasesPage" && slug.current == $slug][0]`;
+  *[_type == "customerCasesPage" && ${translatedFieldFragment("slug")} == $slug][0]{
+    ...,
+    "language": $language,
+    "slug": ${translatedFieldFragment("slug")},
+    "basicTitle": ${translatedFieldFragment("basicTitle")},
+    "seo": ${translatedFieldFragment("seo")} {
```
mathiazom, line 42: If we don't include SEO translation in this PR, this query must be changed from `${translatedFieldFragment("seo")}` to just `seo`

Might be easier to just include the commit from #779

mathiazom: If you include translation, you can also reuse the `SEO_FRAGMENT` from `studio/lib/queries/pages.ts`, but that's not required 🤗

mathiazom (approved):

Issue #733, opened 2024-10-08
Full path translation


mathiazom:
Currently, `languageMiddleware.ts` only supports simple slugs with no nesting (e.g. `/cases`). We should also be able to nest slugs and still get translations (e.g. `/cases/big-corp` with translation `/kunder/korperud`).

mathiazom:
#771 is needed to support this with customer cases, which is the only example we have right now

PR #781, opened 2024-10-15
v3 - translate full path

mathiazom:
See #733 

The `useLanguage` client hook has been removed in favor of providing the path translation together with the page data. This also means that the translations are fetched on the server instead of on the client.

In order to allow this, the header component has been moved from the root `layout.tsx` to each page. This is because the `LanguageSwitcher` in `Header` needs access to the page data, which is not available in `layout.tsx`. This approach is similar to the Sanity example project [`demo-course-platform`](https://github.com/sanity-io/demo-course-platform/blob/e8e2a2cf026539a232886500fa9fb2660a40ceb5/web/app/%5Blanguage%5D/%5Bcourse%5D/page.tsx#L76).


src/middlewares/languageMiddleware.ts
```diff
@@ -221,30 +220,63 @@ export async function languageMiddleware(
   }
   const path = request.nextUrl.pathname.replace(/^\//, "").split("/");
   const language = availableLanguages.find(({ id }) => path[0] === id);
+  const defaultLanguageId = (
+    await client.fetch<LanguageObject | null>(DEFAULT_LANGUAGE_QUERY)
+  )?.id;
+  if (defaultLanguageId === undefined) {
+    console.error(
+      "No default language available, language middleware aborted.",
```
anemne, line 228: Vil vi vise denne meldingen i consol?

mathiazom: Dette gjelder en feil/mangel i oppsettet av Sanity, så tenker det er fint å få en tydelig melding. Siden det kjører på server-siden vil det ikke vises noen feilmelding som brukeren ser i nettleseren.

PR #787, opened 2024-10-16
v3 - update customer cases interface

mathiazom:
The interface type for customer cases was not updated when the schema was changed. This was likely because no web changes have been made yet. This PR updates the interface to match the schema. Also defined a base version of the customer case, intended to be used in list views, while the customer case interface includes all fields used on the detail page.

### DIFF
--- a/studio/lib/interfaces/specialPages.ts
+++ b/studio/lib/interfaces/specialPages.ts
@@ -1,7 +1,5 @@
 import { SeoData } from "src/utils/seo";
 
-import { Slug } from "./global";
-
 export interface CustomerCasePage {
   _createdAt: string;
   _id: string;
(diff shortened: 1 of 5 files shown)


studio/lib/interfaces/specialPages.ts
```diff
@@ -10,7 +8,7 @@ export interface CustomerCasePage {
   _updatedAt: string;
   basicTitle: string;
   page: string;
-  slug: Slug;
+  slug: string;
```
anemne, line 11: Have this changed?

mathiazom: Yep. Since the customer case page is field translated, the slug is returned as a string.

PR #788, opened 2024-10-17
v3 - customer case rich text block

mathiazom:
Introduced customer case sections, where we will expand with different block types to dynamically build customer case pages. This PR only includes the rich text block.

### DIFF
--- a/studio/utils/preview.ts
+++ b/studio/utils/preview.ts
@@ -1,27 +1,20 @@
-import { PortableTextBlock } from "sanity";
+import {
+  PortableTextBlock,
+  isPortableTextSpan,
+  isPortableTextTextBlock,
+} from "sanity";
 
 // Convert rich text content to plaintext preview string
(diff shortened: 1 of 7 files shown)


studio/utils/preview.ts
```diff
@@ -0,0 +1,20 @@
...
+  richText: PortableTextBlock[],
+): string | undefined {
+  const block = richText.find((block) => block._type === "block");
+  if (!isPortableTextTextBlock(block)) {
+    return undefined;
+  }
+  return block.children
+    .filter((child) => isPortableTextSpan(child))
+    .map((span) => span.text)
+    .join("");
```
anemne, line 19: Should we limit the amount of words that is shown in this richtextpreview? and end it with  ... if there is more text?

anemne (approved):

studio/utils/preview.ts
anemne: Great!

PR #789, opened 2024-10-17
v3 - customer cases link

mathiazom:
Added `Link` to customer case titles in the customer case landing page. 

The user can then easily navigate to individual customer case detail pages (even though they are not implemented yet)

image


anemne (approved):
This work for now, but i guess we should use the redirect components that we already have when design is ready.

PR #791, opened 2024-10-17
v3 - customer case images block

mathiazom:
Images block type to be used in the customer case sections array.

Allows multiple images to be uploaded into the same block.

No other fields are included, like the proposed `wide` toggle. This can be added later.

#### Sections list preview
Shows image alt text (first translation), or number of images if multiple:
image
#### Images block
Alt text (first translation) as preview title
image
#### Individual image
With field-translated alt text
image

### Web

No UI yet, but the fetched data looks something like this:

image

### DIFF
--- a/studio/schemas/fields/media.ts
+++ b/studio/schemas/fields/media.ts
@@ -1,6 +1,8 @@
 import { StringInputProps, defineField } from "sanity";
 
 import { StringInputWithCharacterCount } from "studio/components/stringInputWithCharacterCount/StringInputWithCharacterCount";
+import { isInternationalizedString } from "studio/lib/interfaces/global";
+import { firstTranslation } from "studio/utils/i18n";
 
 export enum ImageAlignment {
   Left = "left",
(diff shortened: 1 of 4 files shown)


studio/schemas/fields/media.ts
```diff
@@ -38,6 +48,31 @@ const image = defineField({
   fields: [imageAltField],
 });
 
+export const internationalizedImage = defineField({
```
anemne, line 51: I think it is a bit strange to name it internalinalizedImage

mathiazom: Yea, I guess the other alternative I had was `imageWithInternationalizedAlt`, which was not great either 😅 

The main point was to not replace the original image, to not break other parts of the project. So do you have a suggestion for a different name?

anemne (approved):

studio/schemas/fields/media.ts
anemne: Yeah, I agree that InternationalizedImage is better. I can't think of a better name, so I guess it's fine.

PR #802, opened 2024-10-22
V3 - refactor text component

anemne:
This PR refactors the text components and style names following the switch to the Britti font. The previous implementation included redundant text styles and components.


trulshj (approved):
Makes sense to me!

PR #803, opened 2024-10-22
V3-hotfix-imagerendering

anemne:
The customer case collapses because the system attempts to map image blocks that contain no images. This issue occurs when image data is missing or not properly handled.


trulshj (approved):

trulshj:
This one also takes care of same issue + more https://github.com/varianter/variant.no/pull/801

PR #820, opened 2024-10-25
V3 customer case UI - top section

anemne:
This PR includes updates to the top section of customer cases, aligned with the finalized design. 

Screenshot 2024-10-25 at 14 41 20


src/components/customerCases/customerCase/customerCase.module.css
```diff
@@ -8,33 +8,75 @@
 .content {
   display: flex;
   flex-direction: column;
-  max-width: 1200px;
```
mathiazom, line 11: Why remove the max-width? Makes it pretty hard to read

image

src/components/text/text.module.css
```diff
@@ -60,7 +60,9 @@
   }
 }
 
-/* TODO: we need to add sizes og h4-h6 */
+/* .h4 */
+/* .h5 */
+/* .h6 */
```
mathiazom, line 65: Should these be defined later? Some TODO comments would be nice

studioShared/schemas/documents/customerCase.ts
```diff
@@ -32,7 +32,29 @@ const customerCase = defineType({
       type: "internationalizedArrayText",
       title: "Description",
       description:
-        "Short paragraph displayed at the top of the customer case page",
+        "Short paragraph displayed at the top of the customer case page. Description can not be more than 50 words.",
```
mathiazom, line 35: Why word count instead of characters?

studioShared/lib/queries/customerCases.ts
```diff
@@ -36,7 +36,9 @@ export const CUSTOMER_CASE_QUERY = groq`
       "name": ${translatedFieldFragment("name")},
       "duration": ${translatedFieldFragment("duration")},
       "sector": ${translatedFieldFragment("sector")},
-      "delivery": ${translatedFieldFragment("delivery")},
+      "deliveries": delivery[] {
+        "delivery": ${translatedFieldFragment("delivery")},
+      },
```
mathiazom, line 41: We should be consistent with "deliveries"/"delivery" in schema and GROQ response. I vote for "deliveries" everywhere for the array, and "delivery" for the string value of each array item.

src/styles/global.css
```diff
@@ -40,9 +40,13 @@ html {
 body {
   margin: 0;
   padding: 0;
+  overflow-x: hidden;
```
mathiazom, line 43: Why is this needed? If there is accidental overflowing somewhere, we should fix it there instead.

studioShared/schemas/fields/customerCaseProjectInfo.ts
```diff
@@ -31,7 +34,36 @@ export const customerCaseProjectInfo = defineField({
...
+              name: "delivery",
+              title: "Delivery",
+              type: "internationalizedArrayString",
+            },
+          ],
+          preview: {
+            select: {
+              item: "delivery",
+            },
+            prepare({ item }) {
```
mathiazom, line 54: `item` is very generic, how about just naming it `delivery`?

src/components/customerCases/customerCase/customerCase.module.css
```diff
@@ -8,33 +8,75 @@
...
+
+  @media (max-width: 1024px) {
+    max-width: 100%;
+  }
 }
 
 .projectInfo {
   display: flex;
+  justify-content: space-between;
+  align-items: center;
```
mathiazom, line 58: Project info should not be center-aligned (only has effect on mobile)

image

src/components/customerCases/customerCases.module.css
```diff
@@ -9,8 +9,6 @@
   display: flex;
   flex-direction: column;
   gap: 5rem;
-  max-width: 1200px;
```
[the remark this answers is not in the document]
anemne: If we set a max-width on this section it will become much more narrow than what's in the sketches. I am, however, using a wide screen here and we might want to set a max-width. I'll ask Julia or OP. 

with max width: 
Screenshot 2024-10-28 at 10 38 14

In sketches: 
Screenshot 2024-10-28 at 10 39 17

src/styles/global.css
anemne: You right! I think i forgot to remove this while testing some stuff. I'll fix it now.

src/components/customerCases/customerCases.module.css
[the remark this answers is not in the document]
anemne: Max-width will be 1400px

src/components/customerCases/customerCase/customerCase.module.css
mathiazom, line 1: Since the header is changed, we should reduce the top margin for `.wrapper` (currently at `8rem`), e.g. `4rem`. Could also decrease it further on mobile (e.g. `2rem`).

src/styles/global.css
```diff
@@ -43,6 +43,9 @@ body {
   width: 100vw;
   background-color: var(--primary-bg);
 }
+main {
+  padding: 1rem;
```
mathiazom, line 47: Do we really need this at the global level? This limits what we can do further down (e.g. display image at full page width).

studioShared/schemas/documents/customerCase.ts
```diff
@@ -32,7 +32,29 @@ const customerCase = defineType({
...
+        "Short paragraph displayed at the top of the customer case page. Description can not be more than 300 characters.",
+      validation: (rule) =>
+        rule.custom<{ value: string; _type: string; _key: string }[]>(
+          (value) => {
+            if (!value) return true;
+            if (value) {
+              const invalidItems = value.filter(
+                (item) =>
+                  typeof item.value === "string" &&
+                  item.value.length - 1 >= 300,
```
mathiazom, line 44:
```suggestion
                  item.value.length > 300,
```

studioShared/schemas/documents/customerCase.ts
```diff
@@ -32,7 +32,29 @@ const customerCase = defineType({
...
+              );
+
+              if (invalidItems.length) {
```
mathiazom, line 47: Same result, but more readable imo (less js "falsy" magic)
```suggestion
              if (invalidItems.length > 0) {
```

mathiazom (approved):
😎

Issue #815, opened 2024-10-25
Employees in project - UI and logic


anemne:
Each customer case should include a list of employees who have contributed to the project. We need to create mini-profile cards for each employee, linking them to the case. These profile cards should feature essential details (e.g., name, role, photo) and should be displayed as shown in the images below: 

![Image](https://github.com/user-attachments/assets/0755f42a-4a3e-4571-bf2f-278b2523c6a1)
![Image](https://github.com/user-attachments/assets/02f911fc-b27d-4dac-9fa0-b70e6c08e337)

mathiazom:
Should probably specify employees by email (or other readable and sufficiently unique identifier), and then fetch details (name, role, photo) from Chewbacca.

mathiazom:
Closed by #831

PR #831, opened 2024-10-28
v3 - consultants in project

mathiazom:
See #815 

Shows the employee cards for each employee in the project consultants list. Data fetched from Chewbacca.

The project consultants field should now contain the employee emails.

image

#### Desktop

image

#### Mobile
image


studioShared/schemas/fields/customerCaseProjectInfo.ts
```diff
@@ -36,9 +36,10 @@ export const customerCaseProjectInfo = defineField({
     defineField({
       // TODO: We should be able to select the consultants from a list
       name: "consultants",
-      description: "The consultants for the project",
+      description:
```
anemne, line 39: Do we need to use @variant.no or could we only use "amn"?

src/components/customerCases/customerCase/CustomerCase.tsx
```diff
@@ -54,12 +63,14 @@ export default function CustomerCase({ customerCase }: CustomerCaseProps) {
                 {customerCase.projectInfo.delivery}
               </Text>
             </div>
-            <div className={styles.projectInfoItem}>
-              <Text>Konsulenter</Text>
-              <Text className={styles.projectInfoItemValue}>
-                {customerCase.projectInfo.consultants.join(", ")}
-              </Text>
-            </div>
+            {consultantsResult.ok && (
```
anemne, line 66: isn't it enough to write consultantsResult?

studioShared/schemas/fields/customerCaseProjectInfo.ts
mathiazom: We are not sure enough that they are unique across countries (e.g. mom@variant.no and mom@variant.se could be two different people). So going for the safer option now. The user experience may be a little worse, but that could be fixed with a dropdown/search solution (where the internal value does not matter to the user)

src/components/customerCases/customerCase/CustomerCase.tsx
```diff
@@ -91,6 +102,50 @@ export default function CustomerCase({ customerCase }: CustomerCaseProps) {
...
+                          alt={consultant.name}
+                          objectFit="cover"
+                          fill={true}
+                        />
+                      </div>
+                      <div className={styles.consultantInfo}>
+                        <p className={styles.consultantName}>
+                          {consultant.name}
+                        </p>
+                        {consultant.officeName && (
```
anemne, line 127: I think it's nice to have office location, but we should probably ask designers

src/components/customerCases/customerCase/CustomerCase.tsx
mathiazom: Are you thinking just `consultantsResult && ...`?

This is using the `Result` type, so `consultantsResult` will always be an object. So checking if `consultantsResult` is truthy would always return true. The proper check with `Result` is to check if `.ok` is true.

anemne (approved):
I think this looks good!

PR #835, opened 2024-10-29
v3 - images block as row

mathiazom:
Changes the images section to display as a flex row on wide screens. Mobile will still display as a column.

Images are set to share a common height. The amount of cropping will therefore depend on the number of images, and the difference in aspect ratio.

image

image


idamand (approved):
Ser bra ut :)

PR #836, opened 2024-10-29
v3 - full width toggle for images section

mathiazom:
Introduces as "Full Width" toggle on the Images section. This allows the content editor to specify wether or not the images should use the full width of the customer case (1400px), or be limit to the content width (960px)

#### Toggle
image

#### Full Width
image

#### Regular Width
image


idamand (approved):
LGTM :)

Issue #816, opened 2024-10-25
Related project - Sanity schema and UI


anemne:
When creating a customer case, we need to add support for linking to other relevant projects. Users should be able to select and link up to three unique, relevant projects from existing customer cases.

This update requires modifications to both the Sanity schema and the UI. 

![Image](https://github.com/user-attachments/assets/38680474-8b0f-4199-8adc-28de0e2cac25)

mathiazom:
Closed by #837

PR #837, opened 2024-10-29
v3 - featured customer cases

mathiazom:
See #816 

Adds a field to customer cases with a list of customer case reference, used to display a "similar projects" section at the bottom of each customer case. Currently set to a max of three featured cases.

# Sanity reference array field

image

> the reference picker is filtered to prevent a case from featuring itself, as well as featuring a case more than once. This is just to help the user, while the validation function actually enforces these rules.

### Desktop (3 cases)
image

### Mobile
image

### Desktop (2 cases)
image

### Desktop (1 case)
image


anemne (approved):
👌

PR #838, opened 2024-10-30
V3-breadcrumb-menu

anemne:
This PR adds a breadcrumb menu to the header. 

Screenshot 2024-10-30 at 08 18 44


src/components/navigation/header/Header.tsx
```diff
@@ -68,63 +69,103 @@ export const Header = ({
...
+        </header>
+      </FocusOn>
+      {pathname !== "/" &&
+        pathname !== "/" + currentLanguage &&
+        breadCrumb(currentLanguage, pathname)}
+    </>
   );
 };
 
+export const breadCrumb = (currentLanguage: string, pathname: string) => (
```
mathiazom, line 134: Why is this not a regular React JSX component?

```jsx
export const BreadCrumb = ({ currentLanguage, pathname }: BreadCrumbProps) => (
...
```

src/components/navigation/header/Header.tsx
```diff
@@ -68,63 +69,103 @@ export const Header = ({
...
+  <ul className={styles.breadCrumbMenu}>
+    {/* TODO: Home should be updated with translation */}
+    {["Home", ...pathname.split("/").slice(2)].map((e, index, path) => {
```
mathiazom, line 137: `.slice(2)` only works if the language is included in the path. We should check if `currentLanguage` is included in `pathname`, and remove accordingly.

image

mathiazom: Or we can use the `path` array from `page.tsx` instead of `usePathname`

mathiazom:
Ideally, the breadcrumb titles would be based on the page titles from Sanity, not just the uppercased slugs. However, that might be quite complex, so this is probably a good initial version 👍

src/components/navigation/header/header.module.css
```diff
@@ -149,3 +145,43 @@
...
+}
+
+.dotSeparator {
+  font-size: 2.5em;
+}
+
+.lastDotSeparator {
+  font-size: 2.5em;
+  color: var(--secondary-red);
+}
```
mathiazom, line 187:
```suggestion
.dotSeparator, .lastDotSeparator {
  font-size: 2.5em;
}

.lastDotSeparator {
  color: var(--secondary-red);
}
```

src/components/navigation/header/Header.tsx
```diff
@@ -68,63 +69,102 @@ export const Header = ({
...
+              {renderPageCTAs(ctas, false)}
+              <div className={styles.languageSwitcher}>
+                {defaultLanguage && (
+                  <LanguageSwitcher
+                    currentLanguage={currentLanguage}
+                    pathTranslations={pathTranslations}
+                  />
+                )}
+              </div>
+              <button
```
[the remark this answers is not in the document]
mathiazom: This part has not actually been touched, just a change in indentation.

src/components/navigation/header/Header.tsx
anemne: Good catch!

mathiazom:
> Ideally, the breadcrumb titles would be based on the page titles from Sanity, not just the uppercased slugs. However, that might be quite complex, so this is probably a good initial version 👍

Defined an issue for this: #842

src/components/navigation/header/header.module.css
```diff
@@ -149,3 +145,43 @@
   composes: button;
   background: url("/_assets/menu-close.svg") no-repeat 50% 50%;
 }
+
+.breadCrumbMenu {
+  position: relative;
```
mathiazom, line 150: Is this `position` needed?

src/components/navigation/header/header.module.css
```diff
@@ -149,3 +145,43 @@
...
+  color: var(--primary-text);
```
mathiazom, line 151: `--primary-text` does not seem to be defined

src/components/navigation/header/Header.tsx
```diff
@@ -68,60 +74,109 @@ export const Header = ({
...
+        </header>
+      </FocusOn>
+      {pathname !== "/" && pathname !== "/" + currentLanguage && (
+        <BreadCrumb currentLanguage={currentLanguage} pathname={pathname} />
+      )}
+    </>
+  );
+};
+
+export const BreadCrumb = ({ currentLanguage, pathname }: BreadCrumbProps) => {
```
mathiazom, line 139: Maybe move this to its own file?

src/components/navigation/header/header.module.css
anemne: I think with the current jfx structure, we don’t need it. I’ll check and remove it if that’s the case.

src/components/navigation/header/Header.tsx
```diff
@@ -68,60 +74,109 @@ export const Header = ({
...
+              <span
+                className={
+                  index < path.length - 2
+                    ? styles.dotSeparator
+                    : styles.lastDotSeparator
+                }
+              >
+                •
+              </span>
+            )}
```
mathiazom, line 175: Maybe place the dot separator before the current slug, instead of after the previous slug? Probably doesn't matter, but I think it makes more sense that the highlighted dot belongs to the last list item.

anemne: Done

src/components/navigation/breadCrumbMenu/BreadCrumbMenu.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+        return (
+          <li key={index} className={styles.breadCrumb}>
+            <Link className={styles.breadCrumb} href={href}>
+              {index !== 0 && (
+                <span
+                  className={
+                    index < path.length - 1
+                      ? styles.dotSeparator
+                      : styles.lastDotSeparator
```
mathiazom, line 37:
```suggestion
                    isLast
                      ? styles.lastDotSeparator
                      : styles.dotSeparator
```

src/components/navigation/breadCrumbMenu/BreadCrumbMenu.tsx
```diff
@@ -0,0 +1,57 @@
...
+}: BreadCrumbProps) => {
+  return (
+    <ul className={styles.breadCrumbMenu}>
+      {/* TODO: "Hjem" should be updated with translation */}
+      {[
+        "Hjem",
+        ...(pathname.includes("/" + currentLanguage + "/")
+          ? pathname.split("/").slice(2)
+          : pathname.split("/").slice(1)),
+      ].map((e, index, path) => {
```
mathiazom, line 24: Maybe rename `e` to `slug` or `pathSegment`

src/components/navigation/breadCrumbMenu/BreadCrumbMenu.tsx
```diff
@@ -0,0 +1,57 @@
...
+                  •
+                </span>
+              )}
+              <Text
+                className={
+                  isLast ? styles.breadCrumbText : styles.breadCrumbLink
+                }
+                type={isLast ? "labelSemibold" : "desktopLink"}
+              >
+                {e.charAt(0).toUpperCase() + e.slice(1).toLowerCase()}
```
mathiazom, line 49: Just in case 🙃 

```suggestion
                {e.charAt(0).toUpperCase() + (e.length > 1 ? e.slice(1).toLowerCase() : "")}
```

PR #839, opened 2024-10-30
HOTFIX - fix rendering issue

anemne:


mathiazom:
How is `featuredCase` not defined? The type should probably be updated if this check is required.

anemne:
I dont know, this just works!

PR #841, opened 2024-10-30
v3 - customer case split section

mathiazom:
Introduces a new customer case section type: `SplitSection`™

This section allows two other sections to be displayed side-by-side (flex row) inside a customer case.

To allow a section to be placed on the right, with only blank space on the left, the `EmptySection` has been added. This section simply renders an empty `div`.

On narrower screens, the sections are simply rendered as a column, and ignoring empty sections.

Below is an attempt at recreating the sketch from Figma:

splitsection


anemne (approved):
Looks good!

Issue #842, opened 2024-10-30
Proper breadcrumb titles


mathiazom:
Currently, breadcrumb titles are based on slugs. This is fine in some cases, but is less helpful where less descriptive slugs are used. E.g. the employee page, where the employee alias is rather cryptic.

mathiazom:
Closed by #846

PR #846, opened 2024-10-31
v3 - proper breadcrumb titles

mathiazom:
See #842 

Extended page data to include an array of page titles. This is then used to display the actual page titles for each segment in the breadcrumb trail.

Before | After
--- | ---
image | image
image | image
image | image


idamand (approved):
Looks good to me

Issue #827, opened 2024-10-28
Highlighted option in rich text section


mathiazom:
Should introduce an option in the rich text section to section as highlighted

https://www.figma.com/design/zgXtEvibcgxxvdMDBBaICD/Ny-variant.no%2Fse?node-id=1325-8592&t=8C5n7426zLyHekL9-4

![Image](https://github.com/user-attachments/assets/75d4b6b2-7480-40c0-b767-fe269928bd74)

mathiazom:
Closed by #848

PR #848, opened 2024-10-31
v3 - highlighted rich text section

mathiazom:
See #827 

Added option to highlight a rich text section. When toggled, the rich text will be wrapped in a yellow frame.

**Studio**

image
image

**Customer case**

highlight


idamand (approved):
Looks good :)

Issue #817, opened 2024-10-25
Block - Results


anemne:
We need to add a "Results" block that includes the following elements:

- Title: A heading to introduce the results section.
- Metrics: Up to three sets of numbers with accompanying descriptive text to highlight key outcomes.

![Image](https://github.com/user-attachments/assets/ee22140f-ab37-4d9b-b86d-13448f14494d)

PR #857, opened 2024-11-04
V3 customer case block results

idamand:
Added result block to Customer Case Page sections ref #817. It includes a block title, and a list of up to five results, each result with a field for result and description. This block cannot be used with split sections. 
Screenshot 2024-11-04 at 14 03 00

Screenshot 2024-11-04 at 14 04 09

### DIFF
--- a/src/components/customerCases/customerCase/sections/CustomerCaseSection.tsx
+++ b/src/components/customerCases/customerCase/sections/CustomerCaseSection.tsx
@@ -2,6 +2,7 @@ import { CustomerCaseSection as CustomerCaseSectionObject } from "studioShared/l
 
 import ImageSection from "./image/ImageSection";
 import QuoteBlock from "./quote/QuoteBlock";
+import ResultBlock from "./results/ResultBlock";
 import RichTextSection from "./richText/RichTextSection";
 import SplitSection from "./splitSection/SplitSection";
 
@@ -19,5 +20,7 @@ export function CustomerCaseSection({
       return <QuoteBlock section={section} />;
     case "imageBlock":
       return <ImageSection section={section} />;
+    case "resultBlock":
+      return <ResultBlock section={section} />;
   }
 }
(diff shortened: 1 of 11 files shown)


src/components/customerCases/customerCase/sections/CustomerCaseSection.tsx
```diff
@@ -2,6 +2,7 @@ import { CustomerCaseSection as CustomerCaseSectionObject } from "studioShared/l
 
 import ImageSection from "./image/ImageSection";
 import QuoteBlock from "./quote/QuoteBlock";
+import ResultBlock from "./results/ResultBlock";
```
anemne, line 5: We might consider naming it resultsBlock, as we may add multiple results.

src/components/customerCases/customerCase/sections/results/ResultBlock.tsx
```diff
@@ -0,0 +1,32 @@
...
+          </Text>
+          <div className={styles.resultrow}>
+            {section.resultList?.map((result) => (
+              <div className={styles.results} key={result._key}>
+                <Text type="h2" className={styles.mainresult}>
+                  {result.result}
+                </Text>
+                <Text type="labelRegular">{result.description}</Text>
+              </div>
+            ))}{" "}
```
anemne, line 26:
```suggestion
            ))}
```

src/components/text/text.module.css
```diff
@@ -1,14 +1,17 @@
...
 .h1,
 .h2,
 .h3,
 .h4,
 .h5,
 .h6,
-.bodyXl {
-  color: var(--primary-black);
+.bodyXl) {
+ color: var(--primary-black);
```
anemne, line 11: It might be better to just remove color from text component?

src/components/text/text.module.css
```diff
@@ -24,8 +27,8 @@
 .bodyNormal,
 .bodyBig,
 .list {
-  color: var(--primary-black);
-  font-family: var(--font-britti-sans);
+  /*color: var(--primary-black);
+  font-family: var(--font-britti-sans);*/
```
anemne, line 31: We can remove the color, but we should keep the font. Should not be commented out.

studioShared/lib/queries/customerCases.ts
```diff
@@ -74,6 +74,14 @@ export const CUSTOMER_CASE_QUERY = groq`
           ${BASE_SECTIONS_FRAGMENT}
         }
       },
+      _type == "resultBlock" => {
```
anemne, line 77: As i mentioned above, i think it's more appropriate to name it resultsblock.

PR #861, opened 2024-11-05
Replace richtextblock with plain string component

anemne:
Replace the RichText block with a Text component to allow for greater control over the blocks added to customer cases.

example: 
Screenshot 2024-11-05 at 15 39 19


idamand (approved):
Looks good :)

PR #862, opened 2024-11-06
feat: add i18n support for static texts in components and stuff

trulshj:
Changes the LanguageMiddleware to directly rewrite the request so it can be passed into next-intl's middleware

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -1,0 +1,7 @@
+{
+  "CustomerCase": {
+    "customer": "Customer",
+    "project": "Project",
+    "duration": "Duration"
+  }
+}
--- a/src/app/(main)/[locale]/layout.tsx
+++ b/src/app/(main)/[locale]/layout.tsx
@@ -7,7 +7,7 @@ import { getMessages } from "next-intl/server";
 import Footer from "src/components/navigation/footer/Footer";
 import FooterPreview from "src/components/navigation/footer/FooterPreview";
 import SkipToMain from "src/components/skipToMain/SkipToMain";
-import { routing } from "src/i18n/routing";
+import { Locale, routing } from "src/i18n/routing";
 import { getDraftModeInfo } from "src/utils/draftmode";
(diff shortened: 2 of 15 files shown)


src/app/(main)/[locale]/layout.tsx
```diff
@@ -37,9 +40,15 @@ export default async function Layout({
 }: Readonly<{
   children: React.ReactNode;
   params: {
-    lang: string;
+    locale: string;
   };
 }>) {
+  if (!routing.locales.includes(params.locale as "en" | "no")) {
```
Comment (anemne) [inline src/app/(main)/[locale]/layout.tsx:46]:
```suggestion
  if (!routing.locales.includes(params.locale as "en" | "no" | "se")) {
```

messages/en.json
```diff
@@ -0,0 +1,7 @@
+{
```
anemne, line 1: We should create a file for se.json as well

src/i18n/request.ts
```diff
@@ -0,0 +1,18 @@
+import { getRequestConfig } from "next-intl/server";
+
+import { routing } from "./routing";
+
+export default getRequestConfig(async ({ requestLocale }) => {
+  let locale = await requestLocale;
+
+  if (!locale || !routing.locales.includes(locale as "en" | "no")) {
```
anemne, line 8:
```suggestion
  if (!locale || !routing.locales.includes(locale as "en" | "no" | "en" )) {
```

src/i18n/routing.ts
```diff
@@ -0,0 +1,13 @@
+import { createNavigation } from "next-intl/navigation";
+import { defineRouting } from "next-intl/routing";
+
+export type Locale = (typeof locales)[number];
+const locales = ["en", "no"] as const;
+
+export const routing = defineRouting({
+  locales,
+  defaultLocale: "no",
```
anemne, line 9: I think the default could be in Norwegian for now, but we might want to consider switching to English later

anemne (approved):
Great work👏

PR #868, opened 2024-11-06
V3-translations for static variables

anemne:
This PR rebases static variables and adds translations for the relevant fields.


idamand (approved):
Looks good 🙌

PR #869, opened 2024-11-07
V3 - Delete backButton

anemne:
To clean up the code, I've removed the BackButton component, as it’s unnecessary for our design and somewhat redundant.


trulshj (approved):
Sletting av ubrukt kode er 10/10!

PR #878, opened 2024-11-13
V3 replace logo from sanity 

anemne:
Replace icon sourced from Sanity with local svg. This change is made to make it easier for us when developing and keep up with the changes in design. I've not deleted the schema in Studio as we might want to use it for SEO.


idamand (approved):
Looks good :)

PR #908, opened 2024-11-26
✨ Add schemas for job postings

petterhh:
## Summary 
- Display job postings under `Admin`
- Add schemas `jobPosting` and `jobPostings`
  - A job posting consists of `role`, `locations` and `recruiteeAdUrl`. The naming can be discussed (as always!)

## For reviewer
- I have `jobPosting.ts` in `studio/schemas/objects` and `jobPostings.ts` in `studio/schemas/documents/admin`. Correct?

## Screenshots
Screenshot 2024-11-26 at 08 48 55
Screenshot 2024-11-26 at 16 59 10
Screenshot 2024-11-26 at 08 49 28

### DIFF
--- a/studio/schemas/documents/admin/jobPostings.ts
+++ b/studio/schemas/documents/admin/jobPostings.ts
@@ -1,0 +1,21 @@
+import { defineType } from "sanity";
+
+import { jobPostingID } from "studio/schemas/objects/jobPosting";
+
+export const jobPostingsID = "jobPostings";
+
+const jobPostings = defineType({
+  name: jobPostingsID,
+  type: "document",
+  title: "Job Postings",
+  fields: [
+    {
+      name: "jobPostings",
+      title: "Job Postings",
+      type: "array",
+      of: [{ type: jobPostingID }],
+    },
+  ],
+});
+
+export default jobPostings;
--- a/studio/schemas/documents/admin/jobPostings.ts
+++ b/studio/schemas/documents/admin/jobPostings.ts
@@ -10,7 +10,7 @@ const jobPostings = defineType({
   title: "Job Postings",
   fields: [
     {
-      name: "jobPostings",
+      name: "jobPostingsArray",
       title: "Job Postings",
       type: "array",
       of: [{ type: jobPostingID }],
--- a/studio/schemas/objects/jobPosting.ts
+++ b/studio/schemas/objects/jobPosting.ts
@@ -15,6 +15,7 @@ const jobPosting = defineType({
       name: "role",
       type: "internationalizedArrayString",
       description: "The name of the role",
+      validation: (rule) => rule.required().error("Role name is required"),
     },
     {
       title: "Location",
@@ -27,7 +28,15 @@ const jobPosting = defineType({
       title: "Recruitee ad URL",
       name: "recruiteeAdUrl",
       type: "url",
-      description: "URL to Recruitee ad",
+      description:
+        "URL to Recruitee ad. Please  enter the full URL, including 'https://', e.g., 'https://www.example.com'.",
+      validation: (rule) => [
+        rule.required(),
+        rule.uri({
+          scheme: ["http", "https"],
+          allowRelative: false,
+        }),
+      ],
     },
   ],
   preview: {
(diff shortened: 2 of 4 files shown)


studio/schemas/documents/admin/jobPostings.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const jobPostingsID = "jobPostings";
+
+const jobPostings = defineType({
+  name: jobPostingsID,
+  type: "document",
+  title: "Job Postings",
+  fields: [
+    {
+      name: "jobPostings",
```
idamand: Name should be unique, the name of the field is now the same as the name of the document :)

studio/schemas/objects/jobPosting.ts
```diff
@@ -0,0 +1,53 @@
...
+      name: "location",
+      type: "reference",
+      to: [{ type: "companyLocation" }],
+      description: "Where is the role located?",
+    },
+    {
+      title: "Recruitee ad URL",
+      name: "recruiteeAdUrl",
+      type: "url",
+      description: "URL to Recruitee ad",
```
anemne: You should add a rule here:

```
description: "Url to the recrutee post. Please enter full please enter the full URL, including 'https://'. For example, 'https://www.example.com'.",suggestion
validation: (rule) =>
        rule.uri({
          scheme: ["http", "https"],
          allowRelative: false,
        }),
```
I believe we should make all fields required to ensure that job postings are complete and don't lack essential information before being shared

studio/schemas/objects/jobPosting.ts
```diff
@@ -0,0 +1,53 @@
...
+  type: "object",
+  fields: [
+    {
+      title: "Role",
+      name: "role",
+      type: "internationalizedArrayString",
+      description: "The name of the role",
+    },
+    {
+      title: "Location",
```
idamand: You might have already discussed this with either the Ole Petter or Marius, if so please ignore this comment! At the current website we have the option of listing a job to multiple locations if the listing is not location-specific, eg. "Fagleder tjenestedesign", should we also do that here?

studio/schemas/documents/admin/jobPostings.ts
petterhh: Fixed in https://github.com/varianter/variant.no/pull/908/commits/c29246e8d078be00cf34f2504f4464f5920e404c

studio/schemas/objects/jobPosting.ts
petterhh: Fixed in https://github.com/varianter/variant.no/pull/908/commits/c29246e8d078be00cf34f2504f4464f5920e404c

studio/schemas/objects/jobPosting.ts
petterhh: Hmm, I guess we could make the field optional, in which case no location means that it's not tied to a specific location 🤔

idamand: I would send a DM to Marius Krakeli, who is responsible for recruitment in Variant, I believe he has some opinions on whether it should be optional or a list of locations ☺️

petterhh: > I would send a DM to Marius Krakeli, who is responsible for recruitment in Variant, I believe he has some opinions on whether it should be optional or a list of locations ☺️

Done! I'll keep you posted

petterhh: Screenshot 2024-11-26 at 10 36 23

studio/schemas/documents/admin/jobPostings.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const jobPostingsID = "jobPostings";
+
+const jobPostings = defineType({
+  name: jobPostingsID,
+  type: "document",
+  title: "Job Postings",
+  fields: [
+    {
+      name: "jobPostingsArray",
```
petterhh, line 13: I really don't like this name, but as Ane pointed out, `jobPostings` is the name of the type itself, so it might be a bit confusing if `jobPostings` consists of a single field `jobPostings` 😅 Any other name suggestions?

idamand: I think it works, it describes what the array consists of, and the name is only for developers to use when fetching data and such, so I say go for it, unless someone has a better idea 🤷‍♀️

anemne: agreed!

anemne (approved):
Looks good!

PR #910, opened 2024-11-26
checks in seo

anemne:
Empty fields in customer case project deliveries caused the page to crash. The issue arose because the SEO component attempted to map an empty array. This pull request adds checks in the SEO logic to prevent such errors.


idamand (approved):
LGTM :)

PR #914, opened 2024-11-27
V3 fix projectinfo

idamand:
Fixes both the customer page crashing if there was no consultants added to the project and the styling of the project info section. Now the project info section should work no matter how much content is added in the different sections! 

Please do a thorough review on the checks to ensure they catch everything :)

Screenshot 2024-11-27 at 15 00 49
Screenshot 2024-11-27 at 15 01 12


trulshj (approved):
LGTM!

PR #915, opened 2024-11-27
feat: Color Picker for Customer Cases

trulshj:
## A bit cursed, but this allows you to choose a subsection of the title to apply color to

TODO: Validation that the title actually contains the subsection you input

image

## Fancy Color Picker Components!

Possible TODO: Show a preview of title / Badges

image

## How these colors look applied:

image


studioShared/schemas/documents/customerCase.ts
```diff
@@ -39,7 +39,34 @@ const customerCase = defineType({
       type: "internationalizedArrayString",
       title: "Title",
     }),
+    defineField({
```
idamand, line 42: According to the design the colors for the title and badge should be the same, is there any way to validate this or should we add that in the description?

idamand: Or could we use one color picker for the "main" color, and another for the text? 🤔

trulshj: If they should always be the same, then we should only have one field for it.

PR #916, opened 2024-11-28
feat: add styling for new footer component

TrymVeiby:
## Description
Created a new footer component

## Figma
https://www.figma.com/design/zgXtEvibcgxxvdMDBBaICD/Ny-variant.no%2Fse?node-id=3825-28533&m=dev

### Mobile
image

### Tablet
image

### Desktop
image


mikaelbr:
Starting to look good!

should this have max width also?

src/components/navigation/footer/Footer.tsx
```diff
@@ -1,90 +1,156 @@
...
+            )}
+            {navigationData.footer && (
+              <FooterSection title={t("other")}>
+                {renderOtherLinks(navigationData)}
+              </FooterSection>
+            )}
+
+            {soMeData && (
+              <FooterSection title={t("socialMedia")}>
+                {renderSoMe(navigationData, soMeData)}
```
mikaelbr, line 86: Typically this should be a component to utilise react properly, but can be fixed later

idamand: Yep, there are also a few things that can be done to make this more dynamic and utilize the connection to Sanity better, but I would rather get a version of this out now, and then we can make it even better later 👌

Regarding max-width I wasn't quite sure if it should have it or not, as some footers span the width of the page, but it won't be a problem to add it if we see that it's too big in relation to other content

PR #919, opened 2024-11-28
fix: polishing employee page

mikaelbr:
- Makes it load lazily
- Matches design and typography
- Adds proper links
- Changes to be actual grid that supports longer names
- makes it more small-screen friendly

Fixes https://github.com/varianter/variant.no/issues/898

OBS: I haven't checked if the Employee card is used several places and if this change breaks anything outside the listings 😅

Screenshots:

Screenshot 2024-11-28 at 22 39 32
Screenshot 2024-11-28 at 22 39 24

### DIFF
--- a/src/components/employeeCard/EmployeeCard.tsx
+++ b/src/components/employeeCard/EmployeeCard.tsx
@@ -30,7 +30,7 @@ export default function EmployeeCard({
           <div className={styles.employeeImage}>
             <Image
               src={employee.imageUrl}
-              alt=""
+              alt={employee.name}
               style={{ objectFit: "cover" }}
               fill={true}
             />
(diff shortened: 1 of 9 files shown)


src/components/employeeCard/EmployeeCard.tsx
```diff
@@ -30,38 +30,58 @@ export default function EmployeeCard({
           <div className={styles.employeeImage}>
             <Image
               src={employee.imageUrl}
-              alt=""
+              alt={employee.name}
               style={{ objectFit: "cover" }}
               fill={true}
             />
           </div>
         </Link>
```
anemne, line 38: Maybe we should remove the link to the mini-cv for now?

anemne (approved):
Looks good!

src/components/employeeCard/EmployeeCard.tsx
mikaelbr: I think it adds value to be able to link to contact information for a particular person 🤔

anemne: Agreed that we should have links to contact info, but do we need link to the mini-cv page?

mikaelbr: I mean contact page. As in "Contact this person http://localhost:3000/no/folk/mn"

PR #920, opened 2024-11-28
V3 customercase entry module

anemne:
Customer case entry point. Still need some styling and link to customer case. 

When hover: border radius and opacity in image will change: Screenshot 2024-12-09 at 13 35 19

Screenshot 2024-12-09 at 13 35 25

Screenshot 2024-12-09 at 13 35 45

Screenshot 2024-12-09 at 13 37 22

Screenshot 2024-12-09 at 13 37 32


anemne:
Could you @mikaelbr or @trulshj  please take a look at the scaling of this card? I have used way to much time on it now, and can not find a ideal way to solve it😶‍🌫️

mikaelbr:
Yep. Looking at this now. I think it should be similar to some rewrites I do to the image split component

mikaelbr:
Suggestion here @anemne https://github.com/varianter/variant.no/pull/957

messages/se.json
```diff
@@ -14,7 +17,12 @@
     "development": "Utveckling",
     "project_management": "Projektledning",
     "url": "Link",
-    "collaborators": "Samarbeidspartnere"
+    "collaborators": "Samarbeidspartnere",
+    "customer_case_entry": {
+      "image": "bild",
+      "case": "Case",
+      "field": "Field"
```
idamand: Needs a swedish translation of "field"

trulshj (approved):

messages/se.json
anemne: I think we need to get someone speaking Swedish to look at the translations. I'll add something for now, but absolutely not sure if it's correct.

PR #926, opened 2024-11-29
feat: adds split image component and add it to Employee site

### LINKED ISSUE #897: Seksjon "Bli en variant"
Legg til modul med første iterasjon av tekst

Image

mikaelbr:
Adds generic split image module that should work like this:
Can be used on several places, but adds this specific block to employee site.

Fixes https://github.com/varianter/variant.no/issues/897

## Todo later:

- Fix ButtonLink component to match design

Screenshots:

Screenshot 2024-11-29 at 19 18 27
Screenshot 2024-11-29 at 19 18 07
Screenshot 2024-11-29 at 19 20 04

### DIFF
--- a/src/components/sections/image-split/ImageSplit.tsx
+++ b/src/components/sections/image-split/ImageSplit.tsx
@@ -42,9 +42,16 @@ const ImageSplitComponent = ({ section }: ImageSplitProps) => {
           />
         )}
 
-        {section.link && (
+        {section.actions.length > 0 && (
           <div className={styles.textContainer__link}>
-            <LinkButton isSmall link={section.link} />
+            {section.actions.map((action, index) => (
+              <LinkButton
+                key={action._key}
+                type={index === 0 ? "primary" : "secondary"}
+                isSmall
+                link={action}
+              />
+            ))}
           </div>
         )}
       </div>
--- a/src/components/sections/image-split/ImageSplit.tsx
+++ b/src/components/sections/image-split/ImageSplit.tsx
@@ -1,0 +1,42 @@
+import { PortableText, PortableTextReactComponents } from "@portabletext/react";
+
+import { SanityImage } from "src/components/image/SanityImage";
+import CustomLink from "src/components/link/CustomLink";
+import Text from "src/components/text/Text";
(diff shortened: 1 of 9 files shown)


src/components/sections/image-split/ImageSplit.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const ImageSplitComponent = ({ section }: ImageSplitProps) => {
+  const hasImage = section.imageExtended;
+  const alignment = section.imageExtended?.imageAlignment;
+  const showImageToLeft = hasImage && alignment == ImageAlignment.Left;
+  const showImageToRight = hasImage && alignment == ImageAlignment.Right;
+
+  return (
+    <article className={styles.imageSplit}>
+      {showImageToLeft && (
```
mikaelbr, line 21: Could do this css only with order to avoid having image be fore the title in the actual DOM, but would require not to use grid and I used grid to more robustly get it as 50/50 size. 🤔 Might be lazy

mikaelbr:
Legg til secondary link

Screenshot 2024-12-02 at 09 59 28

studio/schemas/objects/sections/imagesplit.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const imageID = "imageSplitSection";
+
+export const imageSplitSection = defineField({
+  name: imageID,
+  title: "Image Split",
+  type: "object",
+  fields: [
+    title,
+    richText,
```
idamand: This might be a problem for another PR, but does richText support different languages? As far as I can see it does not, and we need the content available in norwegian/swedish and english.

And the title should also be internationalized

mikaelbr: We can just make it non-rich-text

idamand: Yeah i think that would be best! It is in line with what we have done on the customer case page, which was to use plain internationalized text to force the correct styling and to prevent content creators from adding their own headings in the text.

idamand (approved):
Looks good :)

PR #929, opened 2024-12-01
feat: adds sanity structure for compensation calculator section

### LINKED ISSUE #896: Modul ansattopplevelse
Image

mikaelbr:
Fixes https://github.com/varianter/variant.no/issues/896

Adds component for showing calculator based on numbers saved in compensation special page. Also extends the compensation special page to have handbook links to be used in the handbook block. This is stored centralised in the compensation page to easier reuse the component a cross different pages without having to add several things manually.

## Todo

- [x] Add translations
- [x] Set default year
- [x] Self-review and see if there are changes I need to do

## Things to follow-up later

- Heading sizes is off (generally)
- Button is wrong (component needs to be updated)
- Haven't added the chevrons but utilising default browser behaviour for number. This can be added in separate PR

<details><summary>Screen shots</summary>
<p>

Screenshot 2024-12-05 at 16 26 05

Screenshot 2024-12-05 at 16 23 43
Screenshot 2024-12-05 at 16 24 53

</p>
</details>

### DIFF
--- a/src/components/linkButton/linkButton.module.css
+++ b/src/components/linkButton/linkButton.module.css
@@ -72,9 +72,13 @@
 
 .secondary {
   /* Todo: this color needs to be changes */
-  color: var(--background-bg-dark);
+  --_borderColor: var(--text-secondary);
+  --_color: var(--background-bg-dark);
+  --_iconColor: var(--text-secondary);
+
+  color: var(--_color);
   background: rgba(255, 255, 255, 0);
-  border: 1px solid var(--text-secondary);
+  border: 1px solid var(--_borderColor);
 
   &:hover {
     background: rgba(31, 31, 31, 0.05);
@@ -85,6 +89,15 @@
   }
 
   &::after {
-    background-color: var(--text-secondary);
+    background-color: var(--_iconColor);
+  }
+}
+.secondary--darkBg {
+  /* @TODO: this color needs to be changed */
+  --_borderColor: var(--background-bg-light-secondary);
+  --_color: currentColor;
+
+  &::after {
+    background-color: currentColor;
   }
 }
--- a/src/components/linkButton/linkButton.module.css
+++ b/src/components/linkButton/linkButton.module.css
@@ -71,7 +71,6 @@
 }
 
 .secondary {
-  /* Todo: this color needs to be changes */
   --_borderColor: var(--text-secondary);
   --_color: var(--background-bg-dark);
   --_iconColor: var(--text-secondary);
--- a/src/components/sections/compensation-calculator/Calculator.tsx
+++ b/src/components/sections/compensation-calculator/Calculator.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { useTranslations } from "next-intl";
 import { use, useState } from "react";
 
 import { calculateSalary } from "src/components/compensations/utils/salary";
@@ -15,14 +16,13 @@ import { Result } from "studio/utils/result";
 
 import styles from "./compensation-calculator.module.css";
 import { Degree, SalaryData } from "./types";
-import { useTranslations } from "next-intl";
 
 type CalculatorProps = {
   localeRes: Promise<LocaleDocument>;
   salariesRes: Promise<Result<SalaryData, unknown>>;
   background: "light" | "dark" | "violet";
   initialDegree: Degree;
-  initialYear: number;
+  initialYear?: number;
 };
 
 export default function Calculator({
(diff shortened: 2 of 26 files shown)


src/components/linkButton/linkButton.module.css
```diff
@@ -72,9 +72,13 @@
 
 .secondary {
   /* Todo: this color needs to be changes */
```
anemne:
```suggestion
```

src/components/sections/compensation-calculator/Calculator.tsx
```diff
@@ -0,0 +1,90 @@
...
+        background={background}
+        selectedId={degree}
+        onValueChange={(selectedOption) =>
+          setDegree(selectedOption.id as Degree)
+        }
+      />
+
+      <div className={styles.inputWrapper}>
+        <InputField
+          label="Året du begynte i relevant arbeid"
```
anemne: I see that you have a to-do task for translations, so I assume you're already aware that this needs to be updated.

src/components/sections/compensation-calculator/CompensationCalculator.tsx
```diff
@@ -0,0 +1,111 @@
...
+export interface CompensationCalculatorProps {
+  language: string;
+  section: CompensationCalculatorSection;
+}
+
+export default async function CompensationCalculator({
+  section,
+  language,
+}: CompensationCalculatorProps) {
+  const salariesRes = getSalaryByYear(2024, language);
```
anemne:
```suggestion
  const salariesRes = getSalaryByYear(new Date().getFullYear(), language);
```

src/components/sections/compensation-calculator/types.ts
```diff
@@ -0,0 +1,9 @@
+import { SalariesByLocation } from "studio/lib/interfaces/compensations";
+
+export type CompensationData = {
+  slug: string;
+  salariesByLocation: SalariesByLocation[];
+};
+export type Degree = "bachelor" | "master";
```
anemne, line 7: Do we need "other" as well?

studio/schemas/documents/compensations.ts
```diff
@@ -39,6 +40,16 @@ const compensations = defineType({
     bonusesByLocation,
     benefitsByLocation,
     salariesByLocation,
+
+    defineField({
+      name: "handbookLinks",
+      title: "Handbook Section Links",
+      description:
+        "Ordered links to sections in handbook to show important information when it comes to benefints. Used by compensation calculator module.",
```
anemne:
```suggestion
        "Ordered links to sections in handbook to show important information when it comes to benefits. Used by compensation calculator module.",
```

src/components/sections/compensation-calculator/CompensationCalculator.tsx
mikaelbr: Yes, this was missing from my TODO. Was doing some handling of min/max/initial as we speak. out in a few

anemne:
I think it looks good so far 😄

src/components/sections/compensation-calculator/CompensationCalculator.tsx
mikaelbr: But actually, this particular line I think we need to do more sophisticated so this is a temporary permanent hardcoded thing. Because we need to fetch "latest" page from Compensation page, to use that. I'll cover it in a separate PR

src/components/sections/compensation-calculator/Calculator.tsx
mikaelbr: Yep. Should be handled now

src/components/sections/compensation-calculator/types.ts
mikaelbr: I know this is a discussion, but it makes it harder to do calculations. I think actually the proper way forward is to instead only communicate "years of relevant experience" and avoid communication on bachelor/masters, but still this is often a requirement from customers so there is a tradeoff here.

anemne: True! I'm not really that keen on going into this discussion again, so I guess this is fine by me.

src/components/linkButton/linkButton.module.css
mikaelbr: Still think the variables here shouldn't be used in this way 🤔 Just match naming in another way

studio/schemas/objects/sections/compensation-calculator.ts
```diff
@@ -0,0 +1,144 @@
...
+        { _key: "no", value: "Ansattopplevelsen" },
+      ],
+    },
+
+    {
+      name: "background",
+      title: "Background",
+      type: "string",
+      description:
+        "Select the whether the calculator should be purple or have dark background.",
```
petterhh:
```suggestion
        "Select whether the calculator should have a purple or dark background.",
```
🤓

petterhh (approved):
Veery minor!

PR #943, opened 2024-12-04
Add Employee Highlight section

### LINKED ISSUE #933: Add people block (aka Marius-block)
Add info block:

- Prefix string
- Title (name) string
- Description richText
- image

Image

petterhh:
## Summary
Add Employee Highlight section

### To reviewer
- What do you think about the "Employee Highlight" name? The "title block" is called "Highlights" in the design, so I just went with it
#### In Sanity Studio
Screenshot 2024-12-04 at 09 47 45
Screenshot 2024-12-04 at 09 47 54

#### Not in Sanity Studio
Screenshot 2024-12-04 at 09 50 18

### DIFF
--- a/src/components/sections/employeeHighlight/employeeHighlight.module.css
+++ b/src/components/sections/employeeHighlight/employeeHighlight.module.css
@@ -16,7 +16,7 @@
 }
 
 .image {
-  min-width: 15rem;
+  min-width: 20rem;
   width: 100%;
   height: auto;
   object-fit: cover;
--- a/src/components/sections/employeeHighlight/employeeHighlight.module.css
+++ b/src/components/sections/employeeHighlight/employeeHighlight.module.css
@@ -14,46 +14,6 @@
   }
 }
 
-.titleSection {
-  display: flex;
-  flex-direction: column;
-  align-items: flex-start;
-  gap: 0;
-}
-
-.title {
-  padding: 0.25rem 1rem;
-  border-radius: 0.25rem 0.25rem 0rem 0rem;
-  background-color: var(--Violet-700);
-  color: var(--text-primary-light);
-}
-
-.nameContainer {
-  display: flex;
-  flex-direction: column;
-  align-items: flex-start;
-  padding: 0.25rem;
-  gap: 0.5rem;
-  border-radius: 0rem 0.25rem 0.25rem 0.25rem;
-  background: var(--Purple-700, #7022d6);
-}
-
-.name {
-  display: flex;
-  flex-direction: column;
-  justify-content: flex-start;
-  align-items: flex-start;
-  padding: 0.75rem 1.25rem 0.625rem 1.25rem;
-  border: 1px solid var(--Violet-700);
-  border-radius: 1.5rem;
-  background-color: var(--text-primary-light);
-}
-
-.description {
-  width: 100%;
-  text-wrap: wrap;
-}
-
 .image {
   border-radius: 0.75rem;
   overflow: hidden;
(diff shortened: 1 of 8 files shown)


studio/schemas/objects/sections/employeeHighlight.ts
```diff
@@ -0,0 +1,56 @@
...
+    },
+    {
+      name: "name",
+      type: "string",
+      title: "Name",
+      description: "The name of the employee.",
+    },
+    {
+      name: "description",
+      type: "internationalizedArrayString",
```
idamand: Perhaps we should use internationalizedArrayText for the description, so the text is in a block rather than one line in Sanity 🤓

petterhh: Fixed in https://github.com/varianter/variant.no/pull/943/commits/dde357498e6bd27d79da33b6545ac66aa5841877

src/components/sections/employeeHighlight/employeeHighlight.module.css
```diff
@@ -0,0 +1,69 @@
+.wrapper {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  padding: 0 2rem;
+  max-width: 64rem;
```
idamand, line 6: I know that the design specifies a width of 64rem, but for the rest of the site we have used 960px for the max-width, as 64rem will be relatively wide on smaller laptop screens 😅

src/components/sections/employeeHighlight/employeeHighlight.module.css
```diff
@@ -0,0 +1,69 @@
...
+  margin: 5rem auto;
+  gap: 1.5rem;
+  color: var(--text-primary);
+
+  @media (max-width: 750px) {
```
idamand: The breakpoint for tablet is 834px and 425px for mobile 📱

src/components/sections/employeeHighlight/employeeHighlight.module.css
```diff
@@ -0,0 +1,69 @@
...
+  min-width: 15rem;
+  width: 100%;
+  height: auto;
+  object-fit: cover;
+
+  & img {
+    border-radius: 0.375rem;
+  }
+
+  @media (max-width: 750px) {
```
idamand: Please use tablet breakpoint of 834px 📱

src/components/sections/employeeHighlight/employeeHighlight.module.css
petterhh: Ole Petter and I decided on 750px because this happens when you go past 750px 😅 
Screenshot 2024-12-05 at 13 40 57

src/components/sections/employeeHighlight/employeeHighlight.module.css
petterhh: I'm seeing multiple pages with max-width set to [1400px](https://github.com/varianter/variant.no/blob/v3-petterhh/employee-highlight/src/components/sections/employees/employees.module.css#L9), [1200px](https://github.com/varianter/variant.no/blob/v3-petterhh/employee-highlight/src/components/employeePage/employeePage.module.css#L10), [1000px](https://github.com/varianter/variant.no/blob/v3-petterhh/employee-highlight/src/components/sections/imageSection/imageSectionComponent.module.css#L3) 🤔

src/components/sections/employeeHighlight/employeeHighlight.module.css
petterhh: But that will obviously depend on the length of the names 😢

src/components/sections/employeeHighlight/employeeHighlight.module.css
idamand: Yeah the maxwidth for the page is different from the maxwidth of some components, so some components we want to take the whole width of the screen e.g. the employee-section on the people-page or some images on the customer case page, whereas other places we want some components to be smaller than the width of the page, like the project info section on the customer case page. Until now I have used a max-width of 960px for all components with width 68rem in Figma, mostly because that was used more in the beginning, but also that the difference between sections that were supposed to be full-width vs. not full-width was almost non-existent on screensizes like 1024px when using 68rem. 

This might be a larger discussion though to ensure consistent use of widths across the page!

src/components/sections/employeeHighlight/employeeHighlight.module.css
idamand: We also should rework our typography a bit, because the font-size should decrease with screensize, I do think that might help in this case actually!

petterhh: > We also should rework our typography a bit, because the font-size should decrease with screensize, I do think that might help in this case actually!

Ah, very true!

idamand: And yep different length of different names is always a pain 😅

src/components/sections/employeeHighlight/employeeHighlight.module.css
idamand: You can leave it for now and we can do a full clean-up either going for 68rem or 960px as we see fit later 🤓

petterhh: Yeeeah, that sounds like a wider discussion indeed 😅

idamand (approved):
🌟

src/components/sections/employeeHighlight/employeeHighlight.module.css
petterhh: Done! https://github.com/varianter/variant.no/pull/943/commits/d62bd413c64ab6ff78a073ea65c58096b0460eb5

src/components/sections/employeeHighlight/employeeHighlight.module.css
petterhh: Done in https://github.com/varianter/variant.no/pull/943/commits/d62bd413c64ab6ff78a073ea65c58096b0460eb5

Issue #830, opened 2024-10-28
Menu


anemne:
![Image](https://github.com/user-attachments/assets/ce1788aa-c3bc-4d8b-bbd5-db5948ddc790)

mikaelbr:
Is this done and should be moved in the board?

PR #944, opened 2024-12-04
V3 update header

### LINKED ISSUE #906: Header design
Design for header

idamand:
The header is now updated with the new design 💫

## Note:
The `contact us` button does not have any functionality yet, it needs to be added before launch. Also, the languageswitch sometimes switches the order of the languages, I could not figure out why.

And the announcement banner needs to be reworked

# Desktop
https://github.com/user-attachments/assets/48c3a948-7a4f-4308-92ce-e04956ad1c51

# Tablet and mobile
https://github.com/user-attachments/assets/6dcf372a-3d71-43db-9b6f-5aaacba0ee7b

# Scrolling 

https://github.com/user-attachments/assets/f42d121c-d2d9-4f8d-bded-b0191909258c

Closes #906 and #830


src/components/navigation/header/Header.tsx
```diff
@@ -92,51 +90,63 @@ export const Header = ({
...
+                  <hr className={styles.divider} />
+                  <div className={styles.mobileButtons}>
+                    {defaultLanguage && (
+                      <LanguageSwitcher
+                        currentLanguage={currentLanguage}
+                        pathTranslations={pathTranslations}
+                      />
+                    )}
+                    <Button size="large" type="primary" background="dark">
+                      <Text type="labelRegular">{t("contact_us")}</Text>
```
anemne, line 144: We should add :mailto in this button

src/components/navigation/header/header.module.css
```diff
@@ -1,52 +1,81 @@
...
 
 .isOpen {
-  height: 100vh;
+  height: 100%;
   display: flex;
   flex-direction: column;
 
   @media (max-width: 1024px) {
-    background: var(--background-bg-dark);
+    background: var(--primary-bg);
```
anemne: Is this a colorvariable that we have?

src/components/navigation/header/header.module.css
```diff
@@ -1,52 +1,81 @@
...
+  position: fixed;
+  top: 0rem;
+  margin: 1.5rem 0;
+  justify-content: center;
   align-items: center;
-  padding: 16px 24px;
+  min-height: 5rem;
+  padding: 0rem 3rem 0rem 3rem;
+  border-radius: 0.75rem;
+  background: rgba(250, 250, 250, 0.75);
```
anemne: Replace with our global variables

anemne (approved):

anemne:
There are some bugs with AccessLint. It does not refresh after it fails, even if you already fixed the access issues. For now, you can ignore the pending check.

src/components/navigation/header/header.module.css
idamand: Because we have lower opacity on the background, we have no global variable for the color of the header, but I can make a new variable for the header background so we can still control it from global.css!

idamand: Fixed in [a401d8e](https://github.com/varianter/variant.no/pull/944/commits/a401d8efa08f1db27892c8091773081b13e964b9)

PR #952, opened 2024-12-06
💄 Align text CSS classes with design

### LINKED ISSUE #950: Make sure that text classes, such as h2, are aligned with the design
See [Slack thread](https://variantas.slack.com/archives/C07TKNC7H46/p1733387283901499)

petterhh:
## Summary
- 💄 Align text styles with design
- 💄 Add `labelLarge`
- 🔥 Remove unused classes `labelSemibold`, `labelBold` and `labelLight`

#### Jobb: Before
Screenshot 2024-12-06 at 11 13 35

#### Jobb: After
Screenshot 2024-12-06 at 11 13 31

#### Folk: Before
Screenshot 2024-12-06 at 11 13 17

#### Folk: After
Screenshot 2024-12-06 at 11 13 13

### DIFF
--- a/src/components/text/text.module.css
+++ b/src/components/text/text.module.css
@@ -91,7 +91,7 @@
 
 .bodySmall {
   font-size: 1rem;
-  font-weight: 300;
+  font-weight: 400;
   line-height: 150%;
 }
 
@@ -110,7 +110,7 @@
 .labelSmall {
   font-size: 0.875rem;
   font-weight: 400;
-  line-height: 1.5rem;
+  line-height: 171.429%;
 }
 
 .labelRegular {
--- a/src/components/text/text.module.css
+++ b/src/components/text/text.module.css
@@ -4,9 +4,7 @@
 }
 
 .labelSmall,
-.labelLight,
 .labelRegular,
-.labelSemibold,
 .labelBold,
 .quoteItalic,
 .quoteNormal,
@@ -22,32 +20,33 @@
 .h1 {
   font-size: 4.5rem;
   font-style: normal;
-  font-weight: 600;
-  line-height: 120%;
+  font-weight: 450;
+  line-height: 111.111%;
 }
 
 .h2 {
-  font-size: 3rem;
-  font-weight: 600;
-  line-height: 120%;
+  font-size: 3.5rem;
+  font-weight: 450;
+  line-height: 114.286%;
 }
 
 .h3 {
-  font-size: 2.125rem;
-  font-weight: 600;
+  font-size: 2rem;
+  font-weight: 450;
+  line-height: 125%;
 }
 
 .h4 {
   font-size: 1.5rem;
   font-style: normal;
-  font-weight: 500;
-  line-height: 120%; /* 1.8rem */
+  font-weight: 450;
+  line-height: 133.3%;
 }
 
 .h5 {
   font-size: 1.25rem;
   font-style: normal;
-  font-weight: 400;
+  font-weight: 450;
   line-height: 120%;
 }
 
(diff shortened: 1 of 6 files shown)


src/components/text/text.module.css
```diff
@@ -71,73 +69,62 @@
...
-  line-height: 120%;
-}
-
-.labelSemibold {
-  font-size: 1rem;
-  font-style: normal;
-  font-weight: 600;
-  line-height: 120%;
+  line-height: 150%;
+  letter-spacing: 0.02rem;
```
mikaelbr, line 121: It's fairly untraditional to add this specifically I think, but it doesn't do any harm if the design actually think that this improves readability etc.

src/components/text/text.module.css
```diff
@@ -22,32 +19,33 @@
...
+  font-weight: 450;
+  line-height: 125%;
 }
 
 .h4 {
   font-size: 1.5rem;
   font-style: normal;
-  font-weight: 500;
-  line-height: 120%; /* 1.8rem */
+  font-weight: 450;
```
mikaelbr, line 41: Have you verified that this actually works as expected? Meaning, that the font included is actually a variable font and supports all types of weight (as this is typically not included in most font types)

mikaelbr: Because I think this should actually use `font-variation-settings: "wght" 450;` 🤔

petterhh: Yup, I'd say it works as expected! For instance, this is weight 450 vs 550:

#### 450
Screenshot 2024-12-06 at 12 41 10

#### 550
Screenshot 2024-12-06 at 12 41 03

src/components/text/text.module.css
petterhh: Well, I just uncritically added it because it was there 😅

src/components/text/text.module.css
mikaelbr: But I think 450 will here show 400, not actually 450. What happens if you replace this line and compare 400 vs 450?:


```suggestion
  font-variation-settings: "wght" 450; 
```
Variable fonts are odd like this. They have settings inside them self that is other than the weight from CSS, so using font-variation-settings instructs the font it self to change.

petterhh: With `font-variation-settings: "wght" 450;`:
Screenshot 2024-12-06 at 12 46 59

vs
`font-weight: 450;`:
Screenshot 2024-12-06 at 12 41 10

mikaelbr: Impossible for me to see the difference, but that might be because the difference is small also 😅

petterhh: Haha yeah, they look quite identical to me

petterhh: If we can't see the difference, I guess we can just go with using `font-weight`?

mikaelbr: Lets just go with it. I'll experiment on my own just because I can't let things go in knowing the details and finding out if it works the way I think it works, but that is my torch to carry on my own.

mikaelbr: cross to carry?

petterhh: cross to bear, I believe!

mikaelbr (approved):
Looks better with less weight! Go for it.

src/components/text/text.module.css
petterhh: > Lets just go with it. I'll experiment on my own just because I can't let things go in knowing the details and finding out if it works the way I think it works, but that is my torch to carry on my own.

Hehehe, keep me posted on your findings!

PR #953, opened 2024-12-06
fix: remove contact box margin

mikaelbr:
Remove the default margin for contact box as it shouldn't implicitly have margin. Might want to see if we should add some layout settings to Sanity for this, but for now I think it is just as good to just remove it.


idamand (approved):
👍

PR #954, opened 2024-12-06
Refactor contact module according to design

idamand:
This refactors the contact box used on the Customer Case page. Using most of the styles and structure used in the contact box from studio. 

# Note:
The text in the component is not changeable in Sanity, only in code. However, the person shown as a contact can be changed in Sanity Studio under Admin -> Company Locations.

# Screenshots
Screenshot 2024-12-06 at 14 22 24

## Mobile
Screenshot 2024-12-06 at 14 24 30


src/components/customerCases/customerCase/contactInformation/ContactInformation.tsx
```diff
@@ -54,21 +54,30 @@ export default async function ContactInformation({
...
+              {t("help")}
+            </Text>
+            <Text type="bodyBig">{t("contact_sale")} </Text>
+          </div>
+          <div className={styles.contactSelectorWrapper}>
+            <Suspense fallback={<EmployeeCardSkeleton background={"light"} />}>
+              <ContactSelector
+                language={language}
+                locations={locationsWithContact}
+                contactByLocation={contactByLocation}
```
mikaelbr, line 73: This shouldn't be awaited, but left a promise to utilise streaming and lazy loading of content (quicker load time)

idamand: What do you mean? Should I not use Suspense?

mikaelbr: Like this

```

const myDataAsPromise = fetchFromSomeWhereWithoutAwait();

// later

<Suspense>
  <MyComponent data={myDataAsPromise} />
</Suspense>


// inside MyComponent

const actualData = use(data); // data refers to promise that was created outside the suspense
```


This will make the data be streamed from the server to the client but not have it block loading

PR #956, opened 2024-12-06
fix: update linkbutton to match figma design

### LINKED ISSUE #936: Reconcile button and linkbutton component
The LinkButton and Button components have different design. Either Button should support A-tag or LinkButton should match the API of Button. I think we should have LinkButton and Button use the same code internally, but expose different Components.

Image

mikaelbr:
Fixes https://github.com/varianter/variant.no/issues/936
Matching button design with component in Figma. And supporting normal link strings.

## Todo later

- Should support all types of icons
- Follow up on some inconsistencies with secondary and background (hover, active etc) that doesn't follow design a cross different views in Figma

## Screenshots

Screenshot 2024-12-06 at 19 26 07
Screenshot 2024-12-06 at 19 26 04
Screenshot 2024-12-06 at 19 25 56

### DIFF
--- a/src/components/navigation/header/Header.tsx
+++ b/src/components/navigation/header/Header.tsx
@@ -7,7 +7,6 @@ import { useEffect, useState } from "react";
 import { FocusOn } from "react-focus-on";
 
 import { defaultLanguage } from "i18n/supportedLanguages";
-import Button from "src/components/buttons/Button";
 import LanguageSwitcher from "src/components/languageSwitcher/LanguageSwitcher";
 import CustomLink from "src/components/link/CustomLink";
 import LinkButton from "src/components/linkButton/LinkButton";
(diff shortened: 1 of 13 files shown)


src/components/navigation/header/Header.tsx
```diff
@@ -102,9 +101,16 @@ export const Header = ({
                     pathTranslations={pathTranslations}
                   />
                 )}
-                <Button size="large" type="secondary" background="light">
-                  <Text type="labelLarge">{t("contact_us")}</Text>
-                </Button>
+
+                {/* THIS SHOULD BE MOVED TO SANITY */}
```
anemne: We could possibly use the contact email field under Admin --> Parent Company

anemne (approved):
lgtm 😎

src/components/navigation/header/Header.tsx
mikaelbr: Should have been separate PR but fixed it here to sneak it in

PR #957, opened 2024-12-08
fix: make customer case responsive 

mikaelbr:
Often when having issues like these (or any issue) it is helpful to re-think the approach and simplify. Here I use just a flexbox with min/max sizing and some breakpoints.

## Screenshots

Screenshot 2024-12-08 at 19 47 56
Screenshot 2024-12-08 at 19 47 41
Screenshot 2024-12-08 at 19 47 35
Screenshot 2024-12-08 at 19 47 26
Screenshot 2024-12-08 at 19 47 21
Screenshot 2024-12-08 at 19 47 15
Screenshot 2024-12-08 at 19 47 09
Screenshot 2024-12-08 at 19 47 04

### DIFF
--- a/src/components/sections/customerCasesEntry/CustomerCasesList.tsx
+++ b/src/components/sections/customerCasesEntry/CustomerCasesList.tsx
@@ -5,11 +5,12 @@ import { useTranslations } from "next-intl";
 import { useState } from "react";
 
 import { SanitySharedImage } from "src/components/image/SanityImage";
-import styles from "src/components/sections/customerCasesEntry/customerCasesEntry.module.css";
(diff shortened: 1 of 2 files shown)


trulshj (approved):
Beautiful!

anemne:
I initially used flex, but switched after discussing it with Jakob. I agree that this approach is much simpler and works perfectly well—or almost perfect! There are still some minor sizing issues, but they should be easy to fix? 
 
Screenshot 2024-12-09 at 11 01 34


Should we consider the number of cases created in relation to the radiobutton group? Perhaps we should limit it to display only 3 cases?

anemne (approved):

PR #958, opened 2024-12-08
feat: adds support for multiple headings and proper sizing for image split component

mikaelbr:
To make image split component more reusable. 

This is not an ideal solution, but is kind of a one size fits almost all. The sanity image component with cover sizing makes this difficult, but I think this is at a bare minimum of what is acceptable before launch. Some images might be cut in some sizes

## Screenshots

Screenshot 2024-12-08 at 21 01 47
Screenshot 2024-12-08 at 21 01 39
Screenshot 2024-12-08 at 21 00 10


trulshj (approved):
image

PR #976, opened 2024-12-09
change translations messages and edit imagesize

anemne:
Fix of this image sizing issue and revert changes in i18n messages. 

Screenshot 2024-12-09 at 15 54 32

### DIFF
--- a/src/components/sections/employeeHighlight/employeeHighlight.module.css
+++ b/src/components/sections/employeeHighlight/employeeHighlight.module.css
@@ -18,7 +18,7 @@
 .image {
   min-width: 20rem;
   width: 100%;
-  height: auto;
+  height: 24rem;
   object-fit: cover;
 
   & img {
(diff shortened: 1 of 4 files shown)


src/components/sections/employeeHighlight/employeeHighlight.module.css
```diff
@@ -18,7 +18,7 @@
 .image {
   min-width: 20rem;
   width: 100%;
-  height: auto;
+  height: 24rem;
```
mikaelbr: Is this intended? Should height ALWAYS be set to 24rem? Why shouldn't it scale?

mikaelbr: In other words: What is this intended to solve by not having it dynamic

anemne: Heheh fixed it now

mikaelbr (approved):
Administasjon er bra!

PR #981, opened 2024-12-10
V3-hero

anemne:
Simplified the existing hero section to contain only a description. 

Screenshot 2024-12-11 at 13 08 32
Screenshot 2024-12-11 at 13 08 45
Screenshot 2024-12-11 at 13 08 58


src/components/sections/hero/hero.module.css
```diff
@@ -1,84 +1,39 @@
...
-    padding: 7.5rem 2rem 5rem 2rem;
+  @media (max-width: 425px) {
+    padding: 0rem;
   }
 }
 
 .shared {
   width: 100%;
-  max-width: 1200px;
+  max-width: 70.125rem;
```
idamand, line 18: We do have a max-content-width variable in global.css, should we perhaps use that one? Or is the width of the hero-section different from that?

idamand (approved):
Looks good to me but check out the comment :)

src/components/sections/hero/hero.module.css
anemne: It was a bit different

petterhh (approved):

PR #988, opened 2024-12-10
fix: employee competency and border radius

mikaelbr:
- Updates new competency fields
- Fallback to show competency even if it doesn't find translation
- Fix so that images are have border radius always

No more Administasjon I'm afraid 😢

Screenshot 2024-12-10 at 20 31 34
Screenshot 2024-12-10 at 20 31 51


idamand (approved):
👌

PR #989, opened 2024-12-10
fix: re-utilise general article content

### LINKED ISSUE #934: Verify Article content for general pages
Article section should have working design and be able to create generic info pages with section content

mikaelbr:
Reutlises Article component as general content page with rich text. This will not support other languages, but I think that is fine for now, as this can be used for info pages that can be linked to etc.

I don't think this is in use as of now and that this will break anything.
Need to verify with design if there are any adjustments that needs to be done. But the requirement here is to just have a way of setting ad-hoc info pages. Example can be temporary Refill page.

Fixes #934 

See screenshot:

Variant av design_ utvikling og strategi · 9 26pm · 12-10

### DIFF
--- a/src/components/richText/richText.module.css
+++ b/src/components/richText/richText.module.css
@@ -17,22 +17,12 @@
 .image {
   overflow: hidden;
   margin: auto;
-  width: 100%;
-  min-width: 150px;
-  max-width: 15rem;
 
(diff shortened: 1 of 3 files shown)


idamand (approved):
Looks good :)

PR #997, opened 2024-12-11
✨ Add openness section

petterhh:
### Summary
- ✨ Add openness section

### Screenshots
Screenshot 2024-12-11 at 15 36 55

#### Sanity Studio
Screenshot 2024-12-11 at 15 38 22

### DIFF
--- a/src/components/sections/openness/openness.module.css
+++ b/src/components/sections/openness/openness.module.css
@@ -1,0 +1,32 @@
+.wrapper {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: flex-start;
+  max-width: var(--max-content-width-large);
+  padding: 0 2rem;
+  margin: 5rem auto;
+  gap: 2rem;
+  color: var(--text-primary);
+}
+
+.image {
+  height: 30rem;
+
+  & img {
+    border-radius: 0.375rem 6rem 0.375rem 0.375rem;
+  }
+}
+
+.title {
+  width: 40rem;
+  color: var(--text-primary);
+}
+
+.content {
+  display: flex;
+  flex-direction: row;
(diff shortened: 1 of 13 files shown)


src/components/sections/openness/openness.module.css
```diff
@@ -0,0 +1,56 @@
+.wrapper {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: flex-start;
+  max-width: var(--max-content-width-large);
+  padding: 0 2rem;
```
idamand: Dont think you need padding here, it will make the width less than the max-width since there is padding put on body instead :)

studio/schemas/objects/sections/openness.ts
```diff
@@ -0,0 +1,42 @@
+import { defineField } from "sanity";
+
+import { titleID } from "studio/schemas/fields/text";
+export const opennessID = "opennessSection";
+
+export const opennessSection = defineField({
```
idamand, line 6: Maybe make everything (title, image, description) required?

idamand (approved):
A few comments but otherwise it looks good 👌

studio/schemas/objects/sections/openness.ts
petterhh: Fixed in https://github.com/varianter/variant.no/pull/997/commits/005770a97b5d15b2ac207e864840680635c05d98

src/components/sections/openness/openness.module.css
petterhh: Fixed in https://github.com/varianter/variant.no/pull/997/commits/005770a97b5d15b2ac207e864840680635c05d98

PR #1011, opened 2024-12-12
💄 Changed from :focus to :focus-visible 

jberglund:
To avoid showing outlines when clicking buttons or checkboxes, only when tabbing/keyboarding


mikaelbr:
Should be added to radio group also

jberglund:
Fann bara en som inte var helt rätt i checkbox.module.css. Ändrat den nu. Kan det stämma eller? @mikaelbr

anemne (approved):

PR #1032, opened 2024-12-12
fix: localtion label for job listings

mikaelbr:
Fixes job listings size

Don't show location when selected location (everyone is the same)

Screenshot 2024-12-12 at 14 38 37
Screenshot 2024-12-12 at 14 38 24
Screenshot 2024-12-12 at 14 38 22


anemne (approved):
👍

PR #1044, opened 2024-12-12
Reduce gap between sections and contact information

### LINKED ISSUE #1016: Openness Section
The gap between the openness and the corresponding contact box is too big

Image

anemne:
fix spacing between openness, learning and generosity section

Screenshot 2024-12-12 at 17 52 32
Screenshot 2024-12-12 at 17 52 47
Screenshot 2024-12-12 at 17 52 59

### DIFF
--- a/src/components/sections/contact-box/contact-box.module.css
+++ b/src/components/sections/contact-box/contact-box.module.css
@@ -3,6 +3,7 @@
   padding: 0;
   width: 100%;
   max-width: var(--max-content-width-large);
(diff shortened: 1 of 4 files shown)


mikaelbr:
This is something that will often have to be compensated for pretty much everywhere this will be used, so it will probably be annoying pretty quick.. 😅 But it's fine for tomorrow and then we can fix permanently as I don't think there is a quick fix for this

mikaelbr (approved):

PR #1068, opened 2024-12-16
refactor: filter out locations with no benefit data

trulshj:


src/components/compensations/CompensationSelector.tsx
```diff
@@ -27,18 +27,23 @@ export default function CompensationSelector({
...
+      id: companyLocation._id,
+      label: companyLocation.companyLocationName,
+    }))
+    .filter((l) => hasBenefits(l.id));
+
   const [selectedLocation, setSelectedLocation] = useState<string>(
-    locations[0]._id,
+    locationOptions[0]?.id,
   );
   const [selectedLocationLabel, setSelectedLocationLabel] = useState<string>(
```
anemne: We dont need this value if we filter the data. It's only used to get the correct city from translations.

anemne (approved):
This is a better solution! Can we add the same to other parts of solution as well?

PR #1071, opened 2024-12-17
Add missing translations for bonus stuff

petterhh:
## Summary
- Add missing translations for bonus stuff
- Change wording from "Velg lokasjon" to "Lokasjon", as we we don't use the "Velg"/"Select" prefix in other filters.
- Change wording from "Historisk bonus" to simply "Bonuser"/"Bonuses", as it's obvious from the years in "Year" column that it's for previous years. Aaaand "historisk bonus" makes it sound like it's the greatest bonus in history 😅  I guess we could use "Tidligere bonuser" or "Årlige bonuser" instead? Thoughts?

### Before
Screenshot 2024-12-17 at 11 58 37

### After
Screenshot 2024-12-17 at 11 58 43

### DIFF
--- a/messages/no.json
+++ b/messages/no.json
@@ -32,7 +32,7 @@
     "contact_sale": "Kontakt salg!",
     "contact_us": "Kontakt oss"
   },
-  "compensation_calculator": {
+  "compensation": {
     "calculator": {
       "formLabel": "Lønnskalkulator",
       "educationInput": "Utdanning",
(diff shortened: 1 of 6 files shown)


messages/no.json
```diff
@@ -44,7 +44,12 @@
       "bachelor": "Bachelor",
       "master": "Master"
     },
-    "test": "test"
+    "bonus": {
+      "location": "Lokasjon",
+      "bonuses": "Bonuser",
```
anemne: Lets name it "tidligere bonuser" of "bonus" in norwegian. I think it sounds a bit strange with bonuser

anemne (approved):

messages/no.json
petterhh: @anemne  Then maybe similarly "previous bonuses" for English?

petterhh: Screenshot 2024-12-18 at 09 16 58

Screenshot 2024-12-18 at 09 17 03

Looks better 👌

anemne: Agreed 👌

PR #1072, opened 2024-12-18
Color to footerwidgets

anemne:
Added a field in Sanity to control the color of the footer widgets. The default color remains yellow, so pages without a specified color will use this default.

Screenshot 2024-12-18 at 10 10 23

Screenshot 2024-12-18 at 09 57 51

Screenshot 2024-12-18 at 09 58 00

Screenshot 2024-12-18 at 09 58 09

### DIFF
--- a/src/app/(main)/[locale]/layout.tsx
+++ b/src/app/(main)/[locale]/layout.tsx
@@ -98,7 +98,7 @@ export default async function Layout({
             companyInfo={initialCompanyInfo.data}
             companyLocations={initialCompanyLocations.data}
             soMeData={initialSoMe.data}
-            footerColorPalette={initialColorPallette.data}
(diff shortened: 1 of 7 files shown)


src/app/(main)/[locale]/layout.tsx
```diff
@@ -54,6 +56,7 @@ export default async function Layout({
     initialSoMe,
     initialLegal,
     initialCompanyLocations,
+    initialColorPallette,
```
Comment (petterhh) [inline src/app/(main)/[locale]/layout.tsx]:
```suggestion
    initialColorPalette,
```
And similarly for the other places where we have pallette instead of palette. Easy mistake to make!

petterhh (approved):
Great idea having this managed in Sanity 🙌 My initial thought was just hardcoding it in an object 😅

PR #1074, opened 2024-12-18
fix: missing props passed to tag causing aria issue

mikaelbr:
Found an issue with props not being passed in tags, causing aria issue on contact box. This fixes that issue. Also reuses a nice helper I found for getting proper tab navigation functionality


petterhh (approved):
👌

anemne (approved):

PR #1089, opened 2025-01-03
Refactor badge to dynamically set text color

### LINKED ISSUE #1024: Customer Case - project info section - badge
Use math to determine text color of the badge based on the chosen title color. Avoid WCAG issues like this:
Image

mathiastorhaug-old:
Updated the badge component to calculate text color based on the background color's luminance.
Also removes the unused sanity badgeText, and refactors surrounding interfaces and props


anemne (approved):
This looks good to me!

PR #1090, opened 2025-01-03
feat: remove hover and icon from external link

TrymVeiby:
## Description
Removed hover effect on custom link external link

## Screenshots
image

## Question for reviewer
Should all external links have this styling or just the link on CustomerCase?

### DIFF
--- a/src/components/link/link.module.css
+++ b/src/components/link/link.module.css
@@ -62,21 +62,6 @@
   @media (min-width: 1024px) {
     font-size: 1.25rem;
   }
-
-  &:hover {
-    gap: 1rem;
-    font-weight: 400;
-  }
-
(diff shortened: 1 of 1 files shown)


src/components/link/link.module.css
```diff
@@ -62,21 +62,6 @@
   @media (min-width: 1024px) {
     font-size: 1.25rem;
   }
-
-  &:hover {
-    gap: 1rem;
-    font-weight: 400;
-  }
-
-  &::after {
```
idamand, line 71: Im a little unsure whether or not the arrow should be removed, it has been used several places in the design so i suspect it is supposed to be there 🤔

idamand (approved):

PR #1112, opened 2025-01-08
Temporary fix: Bug in sitemap

elisekrist:
Commented out customerCasesSitemap due to bug in function that breaks the sitemap


mathiastorhaug-old:
was there no easy way to make customerCasesSitemap work? im just wondering if it would be better to fix it than comment it out hehe 😅

mathiastorhaug-old (approved):

PR #1114, opened 2025-01-08
hover and focus styles for job postings

anemne:
Added styling to job postings. 

hover
Screenshot 2025-01-08 at 12 35 20

focus
Screenshot 2025-01-08 at 12 35 33


mathiastorhaug-old (approved):
nice

PR #1115, opened 2025-01-08
remove bonus table

### LINKED ISSUE #1113: Remove bonus table

mathiastorhaug-old:
It has been decided to remove temporary remove the bonus table, as the information is not correct. We are also unsure if we should have this in the new design. Revert this commit if we want to add it back


anemne (approved):
👌

PR #1116, opened 2025-01-09
📝 Add docs on importing Sanity dataset

petterhh:


docs/import-sanity-dataset.md
```diff
@@ -0,0 +1,17 @@
+### Importing Sanity datasets
+
+Occasionally, we might want to get a fresh dump of the production dataset in Sanity and import it into the staging dataset to avoid having stale data in our local development environments.
+Here's how you can do that!
+
+1. Export the production dataset
+  ```sh
+  sanity dataset export production production.tar.gz
```
mathiastorhaug-old: I think it would be good to explain where to run this. I think it should be possible for someone that has never used sanity before

petterhh: true! should also make a note of how we can differentiate between the two studios

mathiastorhaug-old: yes! It is also not obvious for someone that has never used sanity that there is a difference between studio and manage project. A link or something would be nice

petterhh: @mathiastorhaug Updated in https://github.com/varianter/variant.no/pull/1116/commits/3f82f1393571b9389367ad3e0a5d433e9c18b3db

mathiastorhaug-old (approved):

PR #1147, opened 2025-01-28
fix: Add cache for sanity language query

mathiastorhaug-old:
We had a issue where we would query sanity for supported languages for each field in our schemas. This would result in over 100 api requests on some pages, draining our total api requests

### DIFF
--- a/studio/studioConfig.tsx
+++ b/studio/studioConfig.tsx
@@ -1,10 +1,14 @@
 import { colorInput } from "@sanity/color-input";
 import { documentInternationalization } from "@sanity/document-internationalization";
 import { visionTool } from "@sanity/vision";
+import { SanityClient } from "next-sanity";
 import { WorkspaceOptions } from "sanity";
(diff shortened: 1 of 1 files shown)


studio/studioConfig.tsx
```diff
@@ -17,6 +21,24 @@ import { legalDocumentID } from "./schemas/documents/admin/legalDocuments";
 
 const SUPPORTED_LANGUAGES_QUERY = `*[_type == "languageSettings" && !(_id in path("drafts.*"))].languages[]{id, title}`;
 
+let cachedSupportedLanguages: Promise<Language[]> | null = null;
+//TODO: This might not be the perfect solution, but it works for now.
+function getSupportedLanguages(client: SanityClient) {
+  if (!cachedSupportedLanguages) {
+    console.log("Fetching supported languages");
```
anemne:
```suggestion
```

anemne (approved):
Looks good to me 👌

PR #1162, opened 2025-01-31
Fix handbook urls

anemne:
Updated the query to retrieve correct urls for handbook module.


src/components/sections/compensation-calculator/api.ts
```diff
@@ -17,7 +17,7 @@ export async function getHandbookLinksFromCompensationPage(
     COMPENSATIONS_HANDBOOK_LINKS,
     { language },
     {
-      cache: "force-cache",
+      cache: "no-store",
```
anemne:
```suggestion
      cache: "force-cache",
```

PR #1190, opened 2025-02-14
temp-fix: specifically select calculator id

mikaelbr:
_(don't look ma', I'm doing workarounds)_


mathiastorhaug-old (approved):
Ship it!🚢🚀

PR #1196, opened 2025-02-17
Implement location in benefit

mathiastorhaug-old:
This commit implements location tag in benefits, removing the necessity of duplicating benefits that multiple locations share. It is now possible to select "norge" if a benefit is shared with all of norway.


elisekrist (approved):
Jeg syns det ser riktig og bra ut. Godkjenner!

PR #1252, opened 2025-07-04
Add past events to event page

42tte:
Easiest reviewed commit-by-commit https://variant-no-git-42tte-yolo-variant1.vercel.app/no/events

chore:
- add jetbrain files to gitignore
- add editorconfig, so no manual config of editor should be needed
- make pre-commit githook run fast, not linting unchanged files
- format css
- add stylelint, so css is linted
- add security headers
- add eslint jsx-a11y, to lint for accessibility issues

feature:
- add past events to events page


src/components/eventsPage/eventsPage.tsx
```diff
@@ -1,45 +1,94 @@
...
+  const pastEventPostings = eventPostings.eventPostings.filter((event) => {
+    const eventDate = event.date ? new Date(event.date) : null;
+    if (!eventDate) return true;
+    return eventDate < yesterday;
   });
 
+  const sortedFutureEventPostings = futureEventPostings.sort(sortByDateAsc);
+  const sortedPastEventPostings = pastEventPostings
+    .sort(sortByDateAsc)
+    .reverse();
```
mikaelbr, line 44: Might at some point want to limit the number of past events to show, or structure them differently (with navigating back in time in some sort), but I guess this works for now. Also somewhat nit-picky, I'd say that having `O(2n)` instead of `O(n)` to avoid having an extra function (or parameter to function and avoid the point free notation) should be avoided, although the number of items here is marginal.

src/components/eventsPage/eventsPage.tsx
```diff
@@ -1,45 +1,94 @@
...
-            language={params.locale}
-          />
-        ))}
-      </div>
-    </div>
+    <>
+      <section className={styles.eventSection} aria-labelledby={allEventsId}>
+        <Text type="h1" id={allEventsId}>
+          {t("all_events")}
+        </Text>
```
mikaelbr, line 51: Reading this there is something odd about "All events" and further down "Previous events". It isn't intuitive to me that the events on the top is "upcoming". Especially when past and future events have the exact same styling and is visually similarly prioritised. That being said, I think having them shown adds more value than it confuses, and I know changing style / thinking about this properly will probably mean that it won't happen at first.

mikaelbr (approved):

src/components/eventsPage/eventsPage.tsx
42tte: Agree but I didn't want to change anything on design. 
- There is a visually hidden heading for "Coming events" that could be made non-hidden. 
- Past events could probably have a smaller card for the events than future.
- Past events should probably be paginated.
- My OCD suffers from the difference in sorting between the lists one ASC one DESC. But to me the past list made no sense being ASC, and it actually makes some sense to have the future events ASC. Maybe a third larger size of card could be used for the next event making it possible to sort the future events DESC as well but still getting focus to the next upcoming event.

All this could be addressed when need be.

PR #1257, opened 2025-08-12
fix(sec): update csp with missing value

42tte:
- Adds missing `https://vercel.live/` to `frame-src`
- Change domain list for `script-src` since `nonce` + `'strict-dynamic'` makes it obsolete https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src#strict-dynamic

### DIFF
--- a/src/middlewares/nonceMiddleware.ts
+++ b/src/middlewares/nonceMiddleware.ts
@@ -9,13 +9,13 @@ const contentSecurityPolicy = (nonce: string) => {
   const csp = `
     default-src 'self';
     connect-src 'self' https://variant.innocraft.cloud/ https://g.nav.no/api/v1/;
-    script-src 'self' 'unsafe-inline' 'nonce-${nonce}' 'strict-dynamic' https://variant.innocraft.cloud/ https://vercel.live/ ${
+    script-src 'self' 'unsafe-inline' 'nonce-${nonce}' 'strict-dynamic' ${
       process.env.NODE_ENV !== "production" ? "'unsafe-eval'" : ""
     };
     style-src 'self' 'unsafe-inline';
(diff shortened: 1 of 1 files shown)


src/middlewares/nonceMiddleware.ts
```diff
@@ -9,13 +9,13 @@ const contentSecurityPolicy = (nonce: string) => {
...
     connect-src 'self' https://variant.innocraft.cloud/ https://g.nav.no/api/v1/;
-    script-src 'self' 'unsafe-inline' 'nonce-${nonce}' 'strict-dynamic' https://variant.innocraft.cloud/ https://vercel.live/ ${
+    script-src 'self' 'unsafe-inline' 'nonce-${nonce}' 'strict-dynamic' ${
       process.env.NODE_ENV !== "production" ? "'unsafe-eval'" : ""
     };
     style-src 'self' 'unsafe-inline';
     img-src 'self' data: https://cdn.sanity.io/;
     media-src 'self';
-    frame-src 'self';
+    frame-src 'self' https://vercel.live/;
```
mikaelbr, line 18: I guess vercel.live only ever needed in preview branches? 🤔 That being said it doesn't really do a lot of harm in production build 🤷‍♂️

42tte: Get the error in prod. Is this a sign of something else being wrong?

mikaelbr: Odd. Can't think of why that should happen unless vercel adds something in requests, or someone has activated vercel analytics 🤔 Have to investigate, but avoid the error for now

mikaelbr (approved):

Issue #1175, opened 2025-02-04
Alt text for logos in logo salad


anemne:
Logo salad logos are missing proper alt-text. 

<img width="1019" alt="Image" src="https://github.com/user-attachments/assets/56a9d432-fbd5-4932-841e-1c5d13adcf19" />

PR #1273, opened 2025-09-01
fix: properly use alts for logo salad logos

Klevberg:
Alt text for logos in logo salad
[#1175](https://github.com/varianter/variant.no/issues/1175)

This should fix the problem where the alt text for the logos appear as "[object Object],[object Object]".


TrymVeiby (approved):
Lgtm 👌🏻 Bra catch

PR #1342, opened 2025-09-25
fix lineheights and space distribution between elements

anemne:
## Checklist

This is mainly fixes from the previous fix:  https://github.com/varianter/variant.no/pull/1331

This pr let the title take two lines and instead of truncate in middle of a word it truncates before. 

Screenshot 2025-09-25 at 09 22 36

- [x] Give Sweden a headsup of the changes made so that they can update their website


TrymVeiby (approved):
knall!

PR #1345, opened 2025-09-26
feat(event): Internalized rich text field

TrymVeiby:
## Description
Changes `eventPosting` schema to use internalized rich text field instead of just rich text field. Also added `richTextInternalized` as a field for reuse

## Screenshots
image

## Checklist

### DIFF
--- a/studio/schemas/fields/text.ts
+++ b/studio/schemas/fields/text.ts
@@ -100,6 +100,6 @@ export const richTextInternalizedID = "richTextInternalized";
 
 export const richTextInternalized = defineField({
   name: richTextInternalizedID,
-  title: "Rich text Inernalized",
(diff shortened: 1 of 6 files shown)


studio/schemas/fields/text.ts
```diff
@@ -95,3 +95,11 @@ export const richText = defineField({
   ],
   description: "Add and format rich text content.",
 });
+
+export const richTextInternalizedID = "richTextInternalized";
+
+export const richTextInternalized = defineField({
+  name: richTextInternalizedID,
+  title: "Rich text Inernalized",
```
Tuvald: liten skrivefeil i Internalized bare!

Tuvald (approved):
Lgtm!

PR #1372, opened 2025-10-02
fix: gap between tag elements in eventPostings

anemne:
## Checklist

Fix spacing issues with tag-row as described here:  https://variantas.slack.com/archives/C09CT5VP6V9/p1759396367458469

- [x] Give Sweden a headsup of the changes made so that they can update their website


Tuvald (approved):
Supert!

PR #1398, opened 2025-10-09
feat: rename imageLabel to caption

TrymVeiby:
## Description
Renames imageLabel -> caption
## Issue
#1365 

## Checklist


Tuvald (approved):
Lgtm! Fint at du rydder opp i CSS-filen også:))

PR #1479, opened 2025-11-20
fix: Update relase time and number of releases to show in changelog

anemne:
## Checklist


anemne:
Jeg ser at det kommer en god del eldre commits, men jeg lurer på om det er fordi den starter på versjon 3.0.0, og at den neste gang vil oppdatere releaseversion. Litt usikker.

PR #1489, opened 2025-11-26
Fix: solve issue with fetching blogarticles

anemne:
## Checklist

### DIFF
--- a/src/app/api/blogArticles/route.ts
+++ b/src/app/api/blogArticles/route.ts
@@ -1,5 +1,8 @@
+import { headers } from "next/headers";
 import { NextResponse } from "next/server";
 import Parser from "rss-parser";
+
+import { domainFromHostname } from "src/utils/url";
 
 const parser = new Parser({
   customFields: {
@@ -8,7 +11,11 @@ const parser = new Parser({
 });
 
 export async function GET() {
-  const FEED_URL = "https://blog.variant.no/feed";
+  const domain = domainFromHostname((await headers()).get("host"));
+  const FEED_URL =
+    domain === "Variant.se"
+      ? "https://medium.com/feed/variant-swe"
+      : "https://blog.variant.no/feed";
 
   try {
     const response = await fetch(FEED_URL, {
--- a/src/app/api/blogArticles/route.ts
+++ b/src/app/api/blogArticles/route.ts
@@ -1,8 +1,5 @@
-import { headers } from "next/headers";
 import { NextResponse } from "next/server";
 import Parser from "rss-parser";
-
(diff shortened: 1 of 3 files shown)


src/app/api/blogArticles/route.ts
```diff
@@ -1,17 +1,42 @@
...
   customFields: {
     item: [["content:encoded", "content"]],
   },
 });
 
 export async function GET() {
-  const FEED_URL = "https://blog.variant.no/feed";
+  const domain = domainFromHostname((await headers()).get("host"));
+  const FEED_URL =
+    domain === "Variant.se"
```
robinbaxon: This will not work too good, since domain === "Variant.se" will probably never equal true. Will have to compare case invariant versions of the domain.

src/app/api/blogArticles/route.ts
```diff
@@ -1,17 +1,42 @@
...
+      ? "https://medium.com/feed/variant-swe"
+      : "https://blog.variant.no/feed";
 
   try {
-    const feed = await parser.parseURL(FEED_URL);
+    const response = await fetch(FEED_URL, {
+      headers: {
+        "User-Agent": "VariantWebsite (+https://variant.no)",
```
robinbaxon: Do you maybe want to use the NEXT_PUBLIC_URL here, so the user agent is dynamically set to the client page?
NEXT_PUBLIC_URL will then be https://www.variant.se for sweden, and https://www.variant.no for norway.

src/app/api/blogArticles/route.ts
robinbaxon: I would also suggest pattern-matching here, unless you are VERY sure that "domain" here will always be set to either "variant.no" or "variant.se"? 
Something-psuedo like domain.toLowerCase().Contains("variant.se")

robinbaxon (approved):
Testing curl locally, both norwegian version and swedish version:
🇸🇪 : 
```
baxon@Robins-MacBook-Pro variant.no % curl -s https://medium.com/feed/variant-swe -H "User-Agent: VariantWebsite https://www.variant.se/" -H "Accept: application/rss+xml, application/xml;q=0.9, */*;q=0.8" | grep -c variant-swe
23
```

🇳🇴 :
```
baxon@Robins-MacBook-Pro [variant.no](http://variant.no/) % curl -s -H "User-Agent: https://www.variant.no/" -H "Accept: application/rss+xml, application/xml;q=0.9, */*;q=0.8" https://blog.variant.no/feed | grep -c "https://blog.variant.no/"
23
```

Approved! :)

